PR #4, opened 2016-08-25
Use conda for building (and consequently drop Py 2.6, 3.2, add 3.4, 3.5).

roryyorke:
Build with "conda build" instead of setup.py

Conda only supports Python 2.7, and 3.3 and up, so drop 2.6 and 3.2, and
add 3.4 and 3.5.

Install gfortran for compiling, and nose for tests.

Don't install scipy; get matrix class from numpy instead.  Scipy is
still needed for one example for generalized eigenvalues.

Use the python-control channel for lapack.

Correct license in meta.yaml; the intended license of Slycot seems to GPLv2 (see
gpl-2.0.txt in root).

Change meta.yaml project URL to python-control Slycot project page.

Split out the build script from meta.yaml; this might make life slightly
easier for anyone attempting to build this on Windows.

Removed from meta.yaml some comments that look like they were inherited
from a cookie-cutter template.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,39 @@
+# cargo-culted from python-control's .travis.yml
 language: python
-python:
-  - "2.6"
-  - "2.7"
-  - "3.2"
-  - "3.3"
-# install required system libraries
-before_install:
-  - sudo apt-get update --fix-missing -qq
-  - sudo apt-get build-dep python-scipy -qq
-# command to install dependencies from source
-# note, separating requirements so that travis
-# will get output in less than 5 min and won't
-# terminate, using q to keep build info to a
-# minumum for dependencies
-install:
-  - while [[ 1 ]]; do echo "building deps"; sleep 300; done &
-  - msg_pid=$!
-  - pip install -q coverage
-  - pip install -q coveralls
-  - pip install -q nose scipy
-  - pip install -q scipy
-  - kill $msg_pid
 
-# command to run tests
-script: python runtests.py --coverage
+python:
+  - "2.7"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+
+before_install:
+  - sudo apt-get install gfortran
+  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
+      wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh;
+    else
+      wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+    fi
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  - conda install conda-build
+  - conda config --add channels http://conda.binstar.org/python-control
+  - conda info -a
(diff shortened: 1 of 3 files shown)


cwrowley:
I haven't looked closely at this, but at first glance, this looks great.  Here are some initial thoughts.

It's a minor point, but I don't understand the reason for separating the 2-line build script into a separate file.  If this script were used elsewhere, then sure, but it isn't, so creating another file for this seems like it just makes things more complicated.  The reason explained in the commit message is that it "might make life slightly easier for anyone attempting to build on Windows," which I suppose is possible, but I think we should worry about that once somebody tries to build on Windows.  If somebody is going to build on Windows, I'd think the 2-line build script would be the least of their problems.

Another comment is about the GPL license.  If the license for slycot is GPL, doesn't that conflict with the BSD license for python-control?  Maybe that just means the two packages can never be bundled together?

roryyorke:
On the script separation: I guess the reasoning was half wishful-thinking about Windows, half fig-leaf for my own preferences.  I'll revert this change.

On the license: looks like "current" SLICOT is released under GPLv3 [1].  The first commit of Slycot dates from 5 Feb 2010, under GPLv2-or-later [2].

Hm, I just noticed the ```license='BSD'``` line in python-control/Slycot's setup.py [3], while in HEAD of avventi/master setup.py is GPLv2-only [4].

[1] http://slicot.org/20-site/31-gpl-3-0-standalone
[2] https://github.com/avventi/Slycot/commit/a6c3a3fa9c20e1572c1ae3652abbbb56a8e73060/setup.py
[3] https://github.com/python-control/Slycot/commit/00d5a545345e123dd723141671a5a1fe52629e81/setup.py
[4] https://github.com/avventi/Slycot/commit/eca3b965e54470bed7a44307aca8ac8f35c601ce/setup.py

slivingston:
@roryyorke Thanks for commenting that the license listed in setup.py had been modified. The change was made in commit [f6a27896b6346f7418fb97bf3edb7f3b06ce3b04](https://github.com/python-control/Slycot/commit/f6a27896b6346f7418fb97bf3edb7f3b06ce3b04/setup.py) (the relevant link to the repository of jgoppert/Slycot is <https://github.com/jgoppert/Slycot/commit/f6a27896b6346f7418fb97bf3edb7f3b06ce3b04/setup.py>).

I will revert it back to GPLv2, because we cannot switch the license as such.

slivingston:
@cwrowley Regarding that `control` is under a BSD license whereas Slycot is under GPLv2, we can bundle them together. There is not a conflict between the two licenses [1], [2]. `control` can remain under a BSD license, and Slycot under GPLv2. However, generally source code should not be copied from Slycot into `control` to avoid a situation where some of `control` is under GPL.

[1] https://www.gnu.org/licenses/gpl-faq.html#WhatDoesCompatMean
[2] https://www.gnu.org/licenses/license-list.html#ModifiedBSD

slivingston:
The changes proposed in this pull request are good.

slivingston:
@roryyorke [On line 1 of .travis.yml](https://github.com/python-control/Slycot/blob/a1f66bc2ed881c79c743ea76922e060fb5e637be/.travis.yml#L1), should it be "cargo-culled"?

If so, I can apply the change directly on `master` branch.

roryyorke:
It's a reference to [1] (apologies if I'm missing the joke).

Regarding conda difficulties, which I too have had: we could consider
pinning dependency versions to get off conda's apparently interminable
upgrade treadmill.

[1] https://en.m.wikipedia.org/wiki/Cargo_cult_programming

PR #5, opened 2017-01-11
Fix: a and c matrices must have at least one column in td04ad_r

roryyorke:
Related to https://github.com/python-control/python-control/pull/129#issuecomment-271170675

Previously the a and c matrices submitted to SLICOT's `TD04AD` via Slycot's `td04ad_r` could have zero columns if the submitted transfer functions were constant.  This is reasonable, but f2py (or possibly other numpy component) objected. The fix is to ensure a minimum of 1 column; this doesn't affect the result, since `TD04AD` returns the number of required states, and only the required parts of the a, b, and c matrices are returned.

Unfortunately a similar fix does *not* work for `td04ad_c`; I believe this is due to a bug in SLICOT.  I'll file a separate issue for that.

This does not address the case when there are either no inputs (`m=0`) or outputs (`p=0`); I suspect a similar failure would occur in that case.

### DIFF
--- a/slycot/src/transform.pyf
+++ b/slycot/src/transform.pyf
@@ -1,3 +1,6 @@
+!    -*- f90 -*-
+! Note: the context of this file is case sensitive.
+
 subroutine tb01id(job,n,m,p,maxred,a,lda,b,ldb,c,ldc,scale,info) ! in TB01ID.f
     character intent(in) :: job = 'A'
     integer required,check(n>0) :: n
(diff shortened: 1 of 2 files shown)


slivingston:
I confirmed that your proposed change provides for the correct output. Before merging, I want to understand where the wrapper generated by `f2py` fails when `nr=0`. I will study this before 22 Jan.

roryyorke:
FWIW, this
```fortran
       PROGRAM EMPTY
       REAL X(0) 
       END 
```
doesn't compile on my system (Ubuntu 14.04, compiler "fort77 Version 1.15" (from gcc)); error 
```
Error on line 3: Declaration error for x: nonpositive dimension
```
Further, this fairly authentic looking [Fortran 77 Standard](https://www.fortran.com/F77_std/f77_std.html) says
```
          2.6  Array

          An array is a nonempty sequence of data that has a name
          and a type.  The name of an array is a symbolic name.
```

so it could just be invalid to have an empty array in Fortran 77.

slivingston:
The error that is addressed by this PR is thrown from the function `array_from_pyobj` of f2py, and in particular the message

    ValueError: failed to create intent(cache|hide)|optional array-- must have defined dimensions but got...

is from [line 684 in numpy/f2py/src/fortranobject.c](https://github.com/numpy/numpy/blob/c5e1773f0d77755e21d072eb106b8e51a672bfa8/numpy/f2py/src/fortranobject.c#L684). The conditional block is entered because `count_nonpos` (cf. code around [line 595 of fortranobject.c](https://github.com/numpy/numpy/blob/c5e1773f0d77755e21d072eb106b8e51a672bfa8/numpy/f2py/src/fortranobject.c#L595) detects that the number of columns of A or C is 0.

slivingston:
Regarding the case of no inputs (`m=0`), the same error occurs because the number of columns of D is m, as at [line 239 of transform.pyf](https://github.com/python-control/Slycot/blob/9a098d124cb31c083d26feae746a1f98e26d1a90/slycot/src/transform.pyf#L239) and as can be found in the code that is generated by `f2py`,
```C
  /* Processing variable d */
  d_Dims[0]=max(1,p),d_Dims[1]=m;
  capi_d_intent |= F2PY_INTENT_HIDE|F2PY_INTENT_OUT;
  capi_d_tmp = array_from_pyobj(NPY_DOUBLE,d_Dims,d_Rank,capi_d_intent,Py_None);
```

Regarding the case of no outputs (`p=0`), it succeeds if `m > 0`.

PR #7, opened 2017-01-17
Added new wrappers for AB13BD, AB13DD, AB13ED and AB13FD

painyeph:


painyeph:
From https://github.com/jgoppert/Slycot/pull/6

slivingston:
Thanks for creating these wrappers. Why did you choose the default parameter value 1e-10 for `tol` in the functions `ab13bd` and `ab13dd`? In the case of `ab13bd` in particular, using `tol=0.0` causes the FORTRAN subroutine to use a default as defined [in the documentation](https://github.com/python-control/Slycot/blob/9a098d124cb31c083d26feae746a1f98e26d1a90/slycot/src/AB13BD.f#L120-L127):
```
If the user sets TOL <= 0, then an implicitly computed,
default tolerance, defined by  TOLDEF = N*EPS*NORM(B),
is used instead, where EPS is the machine precision
(see LAPACK Library routine DLAMCH) and NORM(B) denotes
the 1-norm of B.
```

slycot/analysis.py
```diff
@@ -1228,4 +1228,332 @@ def ab09nd(dico,job,equil,n,m,p,A,B,C,D,alpha=None,nr=None,tol1=0,tol2=0,ldwork=
...
+
+    Required arguments:
+        n : input int
+            The order of the matrix A.  n >= 0.
+        A : input rank-2 array('d') with bounds (n,n)
+            The leading n-by-n part of this array must contain the matrix A.
+
+    Optional arguments:
+        tol : Specifies the accuracy with which beta(A) is to be
+              calculated. (See the Numerical Aspects section below.)
```
slivingston: The section Numerical Aspects, which is referenced here, was not also copied from the FORTRAN sources. Can you either write that the documentation of the FORTRAN source code should be consulted for details, or copy the Numerical Aspects section ([lines 124-143 of AB13FD.f](https://github.com/python-control/Slycot/blob/9a098d124cb31c083d26feae746a1f98e26d1a90/slycot/src/AB13FD.f#L124-L143)) into this docstring?

painyeph:
You are right for `ab13bd`, `tol` should be 0. It is my mistake.

For `ab13dd`, `tol = 0` often fails. I don't know if it is better to choose 0 as default value. The [SLICOT example](http://slicot.org/objects/software/shared/doc/AB13DD.html#Example) uses 1e-9.

PR #10, opened 2017-04-18
TB05AD wrapper

rabraker:
This pull request is to address issue [116 in python-control](https://github.com/python-control/python-control/issues/116). It provides a wrapper for the TB05AD slicot routine which computes the frequency response of a statespace system. The routine is mostly useful when computing the frequency response at many frequencies. 

The wrapper implements three different modalities for TB05AD:
1). You pass in general (A, B, C) matrices. The A matrix is balanced, transformed to upper Hessenberg form, and the B, and C matrices are appropriately transformed. This returns the frequency response as well as the transformed (A, B, C) matrices.
2).  The same as (1), except no balancing is performed.
3). You pass matrices which have already been transformed to upper Hessenberg form. 

These three different jobs are wrapped as three different f2py functions in transform.pyf. In python, they are all called via the same function in transform.py 'tb05ad' by setting a different job flag. These are 'AG' (All, general), 'NG' (none, general), 'NH' (none, hessernberg) respectively.

Using job 'NG' (2) then job 'NH' (3) works perfectly as far as I can tell. 

For certain A matrices, job 'AG' (1) gives the wrong the results. I believe this is a bug in TB05AD.f circa line 354. Above that line, I have commented out my proposed change which I believe fixes the issue. But I wanted somebody else to see it fail as well to make sure I'm not crazy :). When doing the balancing, TB05AD calls the lapack routine DGEBAL. From what I can tell, TB05AD misinterprets the permutation information provided by DGEBAL, but this misinterpretation is only a problem when the permutation corresponds to non-symmetric permutation matrix. This is similar to the scipy matrix_balance bug which I reported [here](https://github.com/scipy/scipy/issues/7258). Before this was fixed, TB05AD and linalg.matrix_balance would yield the same transformed set (A, B, C), but you would get the wrong frequency response. 

The failure is documented in the unit tests as a known failure.  

I deviated some in writing the unit tests from what is currently implemented. The motivation was to test the three different routines on the same set of matrices while avoiding copy-and-paste coding. I'm sure there is a better way...

Let me know what you think.

Thanks!

Arnold

### DIFF
--- a/slycot/src/transform.pyf
+++ b/slycot/src/transform.pyf
@@ -136,6 +136,100 @@ subroutine tb04ad_c(rowcol,n,m,p,a,lda,b,ldb,c,ldc,d,ldd,nr,index_bn,dcoeff,lddc
     integer optional :: ldwork = max(1,n*(n+1)+max(n*p+2*n+max(n,p),max(3*p,m)))
     integer intent(out) :: info
 end subroutine tb04ad_c
+subroutine tb05ad_ag(baleig,inita,n,m,p,freq,a,lda,b,ldb,c,ldc,rcond,g,ldg,evre,evim,hinvb,ldhinv,iwork,dwork,ldwork,zwork,lzwork,info)
+  ! The case where A is is a general matrix that should be balanced
+  ! and converted to upper Hessenberg form.
+  fortranname tb05ad
+  character intent(hide) :: baleig = 'A'
+  character intent(hide) :: inita = 'G'
+  integer check(n>0) :: n
+  integer check(m>0) :: m
+  integer check(p>0) :: p
+  complex*16 intent(in) :: freq
+  double precision intent(in,out,copy),dimension(n,n),depend(n) :: a
+  integer intent(hide),depend(a) :: lda=shape(a,0)
+  double precision intent(in,out,copy),dimension(n,m),depend(n,m) :: b
+  integer intent(hide),depend(b) :: ldb=shape(b,0)
+  double precision intent(in,out,copy),dimension(p,n),depend(n,p) :: c
+  integer intent(hide),depend(c) :: ldc=shape(c,0)
+  double precision intent(out) :: rcond
+  complex*16 intent(out),dimension(ldg,m),depend(ldg,m) :: g
+  integer intent(hide),depend(p) :: ldg = p
+  double precision intent(out),dimension(n),depend(n):: evre
+  double precision intent(out),dimension(n),depend(n):: evim
+  complex*16 intent(out),dimension(ldhinv,m),depend(ldhinv,m) :: hinvb
+  integer intent(hide),depend(n) :: ldhinv = n
+  ! cache variables
+  integer intent(hide,cache),dimension(n) :: iwork
+  double precision intent(hide,cache),dimension(ldwork),depend(ldwork) :: dwork
+  integer optional,depend(n) :: ldwork = 2*n
+  complex*16 intent(hide,cache),dimension(lzwork),depend(lzwork) :: zwork
+  integer optional,depend(n) :: lzwork = n*n+2*n
+  integer intent(out) :: info
+end subroutine tb05ad_ag
+subroutine tb05ad_ng(baleig,inita,n,m,p,freq,a,lda,b,ldb,c,ldc,rcond,g,ldg,evre,evim,hinvb,ldhinv,iwork,dwork,ldwork,zwork,lzwork,info)
+  ! The case where A is is a general matrix that should not be
+  ! balanced but is only converted to upper Hessenberg form.
+  fortranname tb05ad
+  character intent(hide) :: baleig = 'N'
+  character intent(hide) :: inita = 'G'
+  integer check(n>0) :: n
+  integer check(m>0) :: m
+  integer check(p>0) :: p
+  complex*16 intent(in) :: freq
+  double precision intent(in,out,copy),dimension(n,n),depend(n) :: a
+  integer intent(hide),depend(a) :: lda=shape(a,0)
+  double precision intent(in,out,copy),dimension(n,m),depend(n,m) :: b
+  integer intent(hide),depend(b) :: ldb=shape(b,0)
+  double precision intent(in,out,copy),dimension(p,n),depend(n,p) :: c
+  integer intent(hide),depend(c) :: ldc=shape(c,0)
+  double precision intent(hide) :: rcond
+  complex*16 intent(out),dimension(ldg,m),depend(ldg,m) :: g
+  integer intent(hide),depend(p) :: ldg = p
+  double precision intent(hide),dimension(n),depend(n):: evre
+  double precision intent(hide),dimension(n),depend(n):: evim
+  complex*16 intent(out),dimension(ldhinv,m),depend(ldhinv,m) :: hinvb
+  integer intent(hide),depend(n) :: ldhinv = n
+  ! cache variables
+  integer intent(hide,cache),dimension(n) :: iwork
+  double precision intent(hide,cache),dimension(ldwork),depend(ldwork) :: dwork
+  integer optional,depend(n) :: ldwork = 2*n
+  complex*16 intent(hide,cache),dimension(lzwork),depend(lzwork) :: zwork
+  integer optional,depend(n) :: lzwork = n*n+2*n
+  integer intent(out) :: info
+end subroutine tb05ad_ng
+
+subroutine tb05ad_nh(baleig,inita,n,m,p,freq,a,lda,b,ldb,c,ldc,rcond,g,ldg,evre,evim,hinvb,ldhinv,iwork,dwork,ldwork,zwork,lzwork,info)
+  ! The case where A is already balanced and hessenberg
+  fortranname tb05ad
+  character intent(hide) :: baleig = 'N'
+  character intent(hide) :: inita = 'H'
+  integer check(n>0) :: n
+  integer check(m>0) :: m
+  integer check(p>0) :: p
+  complex*16 intent(in) :: freq
+  double precision intent(in),dimension(n,n),depend(n) :: a
+  integer intent(hide),depend(a) :: lda=shape(a,0)
+  double precision intent(in),dimension(n,m),depend(n,m) :: b
+  integer intent(hide),depend(b) :: ldb=shape(b,0)
+  double precision intent(in),dimension(p,n),depend(n,p) :: c
+  integer intent(hide),depend(c) :: ldc=shape(c,0)
+  double precision intent(hide) :: rcond
+  complex*16 intent(out),dimension(ldg,m),depend(ldg,m) :: g
+  integer intent(hide),depend(p) :: ldg = p
+  double precision intent(hide),dimension(n),depend(n):: evre
+  double precision intent(hide),dimension(n),depend(n):: evim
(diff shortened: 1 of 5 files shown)


slivingston:
The email address used in the commits begins with `=`. Is that intentional? This manifests as GitHub not recognizing the commits as being associated with your GitHub handle. OK to keep the email address this way if you prefer.

rabraker:
No, it was not intentional. There was an error in my .gitconfig file. According to [this](https://gist.github.com/trey/9588090), I can change the commit email by rebasing. Is that the preferred method?

roryyorke:
See @slivingston's take on [rebasing in pull requests here](https://github.com/python-control/python-control/pull/100#issuecomment-240917790).  In brief, rebasing in PRs is fine.

You can't depend on python-control in Slycot or its tests; that's why Travis CI is failing.

I'd like to do a more thorough review, but it will take some time.

rabraker:
Thanks for the link. 

I didn't know that a tool like travis CL even existed until today. After removing the dependence on python-control and making a couple tweaks to the transform.pyf file, the builds all succeed now.

slycot/transform.py
```diff
@@ -333,6 +333,254 @@ def tb04ad(n,m,p,A,B,C,D,tol1=0.0,tol2=0.0,ldwork=None):
...
+    if job = 'AG':
+    --------------
+      At:  The A matrix which has been both balanced and
+           transformed to upper Hessenberg form. The balancing
+           transforms A according
+                      A1 =   P^-1 * A * P.
+           The transformation to upper Hessenberg form then yields
+                      At = Q^T * (P^-1 * A * P ) * Q.
+           Note that the lower triangle of At is in general not zero.
+           Rather, it contains information on the orthoganal matrix Q
```
roryyorke: orthoganal => orthogonal

slycot/transform.py
```diff
@@ -333,6 +333,254 @@ def tb04ad(n,m,p,A,B,C,D,tol1=0.0,tol2=0.0,ldwork=None):
...
+           used to transform A1 to Hessenberg form. See docs for lappack
+           DGEHRD(): 
+           http://www.netlib.org/lapack/explore-3.1.1-html/dgehrd.f.html
+           However, it does not apparently to contain information on P, the
```
roryyorke: "apparently to contain" =>"apparently contain" ?

slycot/transform.py
```diff
@@ -333,6 +333,254 @@ def tb04ad(n,m,p,A,B,C,D,tol1=0.0,tol2=0.0,ldwork=None):
...
+        if out[-1] < 0:
+            # Conform fortran 1-based argument indexing to
+            # to python zero indexing.
+            error_text = ("The following argument had an illegal value: "
+                          + arg_list[-out[-1]-1])
+            e = ValueError(error_text)
+            e.info = out[-1]
+            raise e
+        if out[-1] == 1:
+            error_text = ("More than 30*"+str(n)+"iterations are required "
```
roryyorke: why `30*str(n)` ? The docs just say 30 iterations.

slycot/transform.py
```diff
@@ -333,6 +333,254 @@ def tb04ad(n,m,p,A,B,C,D,tol1=0.0,tol2=0.0,ldwork=None):
...
+                       str(C.shape[1]) + "), but expected (" + str(p) +
+                       "," + str(n) + ")")
+        e.info = -10
+        raise e
+
+    # ----------------------------------------------------
+    # Checks done, do computation.
+    if job == 'AG':
+        out = _wrapper.tb05ad_ag(n, m, p, jomega, A, B, C)
+        error_handler(out, arg_list)
```
roryyorke: This is good; it's tidier than the error handling elsewhere in Slycot.  

It might be good to turn your `error_handler` into the constructor for (currently non-existant) `SlycotValueError`, inherited from `ValueError`, that handle alls the re-indexing, etc., in one place.  However, that's for another PR.

slycot/transform.py
```diff
@@ -333,6 +333,254 @@ def tb04ad(n,m,p,A,B,C,D,tol1=0.0,tol2=0.0,ldwork=None):
     kdcoef = max(index)+1
     return A[:Nr,:Nr],B[:Nr,:m],C[:p,:Nr],Nr,index,dcoeff[:porm,:kdcoef],ucoeff[:porm,:porp,:kdcoef]
 
+
+def tb05ad(n, m, p, jomega, A, B, C, job='NG'):
+    """At, Bt, Ct, g_jw, rcond, ev, hinvb = tb05ad_a(n, m, p, jomega, A, B, C):
```
roryyorke: This doc string isn't right: what is returned depends on job (which isn't listed as an argument).

I'm not sure if we (Slycot and python-control) have a good convention for different return forms.  Numpy appears to just omit the return list in the header (e.g., [numpy.linalg.svd](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.svd.html)) and give details below.

`tb05ad_a` => `tb05ad`.

slycot/transform.py
```diff
@@ -333,6 +333,254 @@ def tb04ad(n,m,p,A,B,C,D,tol1=0.0,tol2=0.0,ldwork=None):
...
+    # Fortran function prototype:
+    # TB05AD(baleig,inita,n,m,p,freq,a,lda,b,ldb,c,ldc,rcond,g,ldg,evre,evim,hinvb,ldhinv,
+    # iwork,dwork,ldwork,zwork,lzwork,info)
+
+    # Sanity check on matrix dimensions
+    if A.shape != (n, n):
+        e = ValueError("The shape of A is (" + str(A.shape[0]) + "," +
+                       str(A.shape[1]) + "), but expected (" + str(n) +
+                       "," + str(n) + ")")
+        e.info = -6
```
roryyorke: I'm not sure it's worth faking the `info` attribute.  You haven't done it for the "Unrecognized job" error at the end (where it's not clear what it would be...).

slycot/tests/test.py
```diff
@@ -64,6 +64,111 @@ def test_td04ad_static(self):
...
+                             [ 0.,  0.],
+                             [ 0.,  0.]]),
+                   'C': np.array([[ 0.,  1.,  1.,  0.],
+                             [ 0.,  1.,  0.,  1.],
+                             [ 0.,  1.,  1.,  1.]])}
+
+n = 20
+p = 10
+m = 14
+np.random.seed(40)
```
roryyorke: I think I forgot to click a button somewhere around here when doing the review.

I also think this test case (class? whatever) should be in it's own file (`test_tb05ad.py`).

rabraker:
Thanks for the review. I have quite a few deadlines coming in the next couple of weeks, so it may be a bit before I can get to it though.

slycot/src/transform.pyf
```diff
@@ -135,7 +135,97 @@ subroutine tb04ad_c(rowcol,n,m,p,a,lda,b,ldb,c,ldc,d,ldd,nr,index_bn,dcoeff,lddc
     double precision intent(hide,cache),dimension(ldwork) :: dwork
     integer optional :: ldwork = max(1,n*(n+1)+max(n*p+2*n+max(n,p),max(3*p,m)))
     integer intent(out) :: info
-end subroutine tb04ad_c
+  end subroutine tb04ad_c
+subroutine tb05ad_ag(baleig,inita,n,m,p,freq,a,lda,b,ldb,c,ldc,rcond,g,ldg,evre,evim,hinvb,ldhinv,iwork,dwork,ldwork,zwork,lzwork,info)
+  fortranname tb05ad
```
rabraker: After reading the documentation again I do think 'C' could be useful. The difference between 'N' and 'C' is that 'C' computes RCOND (an estimate of the reciprocal of the condition number of H), 'N' does not. Here, 
H = (freq*I - A)
so if A has modes with zero or very little damping H will become poorly conditioned for frequencies near those modes.  So 'C' could be useful in order to throw a warning that the calculation at a particular frequency may be inaccurate if the condition number is really bad. You could then either just throw that point away or use the information to adjust the points you're evaluating at. 

I'd be willing to add option 'C' at some point in the future if there is interest in that.

The difference between 'A' and 'B' is that 'A' calculates eigenvalues, and RCOND while 'B' only computes eigenvalues. Both balance matrix A. It is my understanding that these two options are only to be used on the "initial" call to TB05AD. In that sense, I don't think its worth adding a different wrapper just to avoid estimating the condition number in a single call.

slycot/tests/test.py
rabraker: I thought so too and now it is.

slycot/tests/test.py
```diff
@@ -64,6 +64,111 @@ def test_td04ad_static(self):
...
+                  'C':  np.random.rand(p, n)}
+
+
+class test_tb05ad(unittest.TestCase):
+
+    def setUp(self):
+        pass
+
+    def test_tb05ad_ng(self):
+        for key in CASES:
```
rabraker: No problem. I had read on the [nump site](https://github.com/numpy/numpy/blob/master/doc/TESTS.rst.txt#writing-your-own-tests) that test functions shouldn't have docstrings. I thought that was odd.

slycot/transform.py
```diff
@@ -333,6 +333,254 @@ def tb04ad(n,m,p,A,B,C,D,tol1=0.0,tol2=0.0,ldwork=None):
...
+            On entry, this array must contain the input/state matrix B.
+
+
+      C :   double precision array, dimension (p,n)
+            On entry, of this array must contain the state/output matrix C.
+
+
+      job : string, 'AG', 'NG', or 'NH'
+            If job = 'AG' ('all', 'general matrix', the A matrix is
+            first balanced via lapack routine DGEBAL with an orthoganal
```
rabraker: Good point. It's not like I'm documenting every other lapack function that gets called.

rabraker:
I believe this have addressed all your comments. I made a few more minor changes to the docstring for tb05ad in transform.py that fixed a few other typos and made the wording between 'job' descriptions more consistent. 

I can rebase all of it into a single commit, but I thought having them separated might be useful for the moment.

slycot/tests/test.py
roryyorke: I wasn't aware of this recommendation, and I've never really thought about the usefulness of test method docstrings, so I suppose I was expressing a personal preference---sorry.

I don't think there's consensus on whether or not to have such docstrings (see, e.g., [this Stackoverflow question](http://stackoverflow.com/q/1726622/1008142)).  python-control has them in some tests, and not in others, so unfortunately we can't get guidance there.

There's a [nose plug-in](https://github.com/MarechJ/nosenodocstrings) that can prevent the behaviour objected to in that numpy reference.

slycot/tests/test.py
```diff
@@ -64,6 +64,7 @@ def test_td04ad_static(self):
             nr,a,b,c,d = transform.td04ad(rc,nin,nout,index,den,num)
 
 
+
```
roryyorke: a nitpick: please remove this new newline, then this file won't be in the changeset at all.

rabraker:
I've removed that newline as requested. I should be able to file a bug report for TB05AD.f in the next few days.

rabraker:
A [Bug report](https://github.com/python-control/Slycot/issues/11) has been submitted, as requested.

roryyorke:
OK, thanks for all the fixes.  This looks good to me.  @slivingston, what do you think?

roryyorke:
Last thing: could you please squash this to one commit? I have push rights now, will merge to master when that's done.

rabraker:
It's all squashed now.

PR #12, opened 2017-06-01
preliminary results from f2c

autodrive:
I managed to 
* find [f2c](https://github.com/barak/f2c) 
* build on a virtual Linux box
* run over slycot *.f files

Here are the preliminary results.  (and I hope I do not make anyone unhappy)  

I think 
* some tests are (more than) necessary : at least comparing results from gfortran with gcc or clang
* may need to reorganize e.g. separating .c files from .f files into a separate folder

Let me know your opinion.  Has anyone tried this before? (Please understand that I am not 100% familiar with Sli(y)cot, Linux, conda, ...)

### DIFF
--- a/slycot/setup.py
+++ b/slycot/setup.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 from __future__ import division, print_function
+
 import glob
 import os
 import sys
@@ -10,11 +11,19 @@ def configuration(parent_package='', top_path=None):
     config = Configuration('slycot', parent_package, top_path)
 
     # can disable building extension to test packaging
-    build_fortran = True
+    build_fortran = False
+    # to test c files from f2c
+    build_c = True
+
+    fortran_sources = []
+    c_sources = []
 
     if build_fortran:
         fortran_sources = glob.glob(
             os.path.join('slycot', 'src', '*.f'))
+    elif build_c:
+        c_sources = glob.glob(
+            os.path.join('slycot', 'src', '*.c'))
     else:
         print('WARNING FORTRAN BUILD DISABLED')
         fortran_sources = []
@@ -26,10 +35,16 @@ def configuration(parent_package='', top_path=None):
     else:
         liblist = ['lapack']
 
-    config.add_extension(
-        name='_wrapper',
-        libraries=liblist,
-        sources=fortran_sources + f2py_sources)
+    if build_c:
+        config.add_extension(
+            name='_wrapper',
+            libraries=liblist,
+            sources=c_sources + f2py_sources)
+    else:
+        config.add_extension(
+            name='_wrapper',
+            libraries=liblist,
+            sources=fortran_sources + f2py_sources)
 
     config.make_config_py()  # installs __config__.py
 
--- a/slycot/src/AB01MD.c
+++ b/slycot/src/AB01MD.c
@@ -1,0 +1,495 @@
+/* AB01MD.f -- translated by f2c (version 20100827).
+   You must link the resulting object file with libf2c:
+	on Microsoft Windows system, link with libf2c.lib;
+	on Linux or Unix systems, link with .../path/to/libf2c.a -lm
+	or, if you install libf2c.a in a standard place, with -lf2c -lm
+	-- in that order, at the end of the command line, as in
+		cc *.o -lf2c -lm
+	Source for libf2c is in /netlib/f2c/libf2c.zip, e.g.,
+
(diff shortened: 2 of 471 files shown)


murrayrm:
A couple of thoughts:

* I think having a C code version will be quite useful for people who don't have easy access to a FORTRAN compiler.

* I would definitely suggest putting the C source code in a separate directory, perhaps something like f2c/.  

* I'm not sure I understand the logic of including the log file (f2c.log) in the repository. 

* Until we get some testing done, I would leave the FORTRAN version as the default.  

* Finally, so that we can keep the two distributions up to date, I wonder whether we shouldn't include the commands for creating the C version of the files in a Makefile at the top level (not sure about that).

autodrive:
Thanks for the comments.

* Will move .c files to a separate folder.
* Reason to add the log file was : there were some warning messages and I assumed that some might want to check which files had which errors.  (I have no problem removing the log file.  If desired, I can prepare a new branch without the log file and ```git push --force```)
* I expect many new commits and tests should follow before any (important) decisions could be discussed.  
* What about closing this pull request for now and submit a new one (or re-open this pull request) when the c version (or python version) is more prepared? (e.g., when it can be built without much problem and pass the majority of the tests.) 
* Another possibility I am thinking is to use a separate fork to develop the C version.

Hope this helps.

roryyorke:
FWIW, the obvious (only?) platform where Fortran compilation is a problem is MS Windows.  My guess of problem areas:
- getting f2c to work with MSVC; it looks like this is possible [1].
- linking against LAPACK and BLAS.  When building for conda, this should link against the Conda-provided MKL.

Ideally the f2c step would be integrated as part of setup.py.

The original SLICOT distribution (see, e.g., [2]) included many examples, which could be used as the starting point for a test suite.  It would be good *anyway* to have a thorough test set.

[1] https://medium.com/@chintanp/using-f2c-on-windows-to-convert-f77-files-to-c-33da825d5de0
[2] http://http.debian.net/debian/pool/main/s/slicot/slicot_5.0+20101122.orig.tar.gz

autodrive:
Thanks again for the comments.

* In the mean while, I could review several [options for wrapping C codes for Python](https://github.com/autodrive/cython_practice) that I could find.  I plan to see if I can automate the [process](https://github.com/autodrive/Slycot/issues/1#issuecomment-306405828).
* Regarding LAPACK/BLAS subroutines that python-control calls, NumPy/SciPy already include [majority](https://github.com/autodrive/Slycot/issues/1#issuecomment-306416663) of them.  If I find no other idea, I might just [locate the source code file](https://github.com/autodrive/lapack) and see if I could use the f2c again.

slivingston:
@autodrive regarding your question
> What about closing this pull request for now and submit a new one (or re-open this pull request) when the c version (or python version) is more prepared?

(quoted from [above](https://github.com/python-control/Slycot/pull/12#issuecomment-305565238)): I recommend that the PR is closed until it is ready for review or discussion (again). E.g., this is motivated because if you push new changes to the open PR, Travis CI will needlessly run tests against it.

slivingston:
for reference, the URL for Travis CI builds of Slycot on PRs: https://travis-ci.org/python-control/Slycot/pull_requests

autodrive:
A progress update:
* I am currently working on [a fork](https://github.com/autodrive/Slycot)
* Trying to setup [an automate-able procedure](https://github.com/autodrive/Slycot/issues/5#issuecomment-357675420). May need a plan B if it does not work on Windows.
* Could find [a discussion](http://icl.cs.utk.edu/lapack-forum/archives/lapack/msg01710.html) on f2c vs LAPACK > 3.2 : LAPACK seemed to switch to Fortran 90 in 2008 but f2c is still more for Fortran 77. (I could see some [error messages](https://github.com/autodrive/Slycot/blob/develop/slycot/src-f2c/f2c.log))
* IMHO these are the important slycot functions : `sb02md_`, `sb02mt_`, `sb03md_`, `tb04ad_`, `td04ad_`, `sg02ad_`, `sg03ad_`, `tb01pd_`, `ab09ad_`, `ab09md_`, `ab09nd_`, `sb01bd_`, `sb02od_`, `sb03od_`, `sb04md_`, `sb04qd_`, `sb10ad_`, `sb10hd_`
* These are the functions that I guess the above functions call but I could not locate .c version after f2c yet : `dlamch_`, `dlamc3_`, `len_trim__`
* `len_trim__` seems an intrinsic function of Fortran90. Maybe I could prepare some 
Hope this helps.

KW

Issue #18, opened 2018-01-13
Why is the repository named Slycot instead of slycot


murrayrm:
Is there a reason that we are using a capitalized repository name?  On some operating systems (eg, MacOS) there can be problems with capitalization and my general preference is to keep everything lower case.  I _think_ I can just rename the repository in github, but not sure what kind of havoc that would cause?

Anyone have thoughts on this?

moorepants:
If you rename the github repo it will automatically redirect and not that URLs are not case sensitive, so it would like not affect anything.

PR #19, opened 2018-01-15
Fix Travis CI build process

murrayrm:
This PR addresses some issues with the Travis CI rules that were getting in the way of successful build tests.  A number of changes have been made here:

* The versions of python have been updated to Python 2.7, 3.5, and 3.6.  Testing in versions 3.3 and 3.3 have been dropped.

* Building is now done in two ways: using setup.py directly and using conda build, depending on the setting of the TEST_CONDA environment variable.

* The LAPACK library is installed from conda-forge instead of binstar.

* In addition to using the (very minimal) unit tests included in `slycot/tests/`, the `python-control` package is now downloaded and tested as well.  This can probably go away when a stronger set of unit tests are available, but since the main use of `slycot` appears to be within `python-control`, it seemed prudent to make sure everything is working there.

* The README documentation has been updated.

* I changed the references to `python-control/Slycot` to `python-control/slycot`, as discussed in issue #18.


roryyorke:
This looks good to me.

jonnor:
Looks like this can be merged? Changes look fine, has one review by @roryyorke, and this seems to be needed for python-control 0.8 release.

PR #24, opened 2018-06-13
Fixes and workarounds for Conda build

roryyorke:
Still in progress, but I have successfully built on Travis with Conda with Python 2.7, 3.5, and 3.6.  The CI build fails with Python 3.6 in the non-Conda build; I'm still investigating that.  See https://travis-ci.org/roryyorke/Slycot/builds/391434012

The main changes:
  - openblas instead of lapack (openblas supplies liblapack and libblas)
  - add -shared to LDFLAGS instead of overwriting it
  - use new meta.yaml layout (not sure if this is critical)

To-do; would appreciate any assistance with these:
  - test on OSX (I can't do this)
  - test on Windows, in particular check I haven't broken @repagh's work on Windows builds
  - investigate test failure

### DIFF
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -1,2 +1,20 @@
+#!/bin/bash
 cd $RECIPE_DIR/..
-LDFLAGS="-shared" FFLAGS="-fPIC" $PYTHON setup.py install
+
+# conda's activate-gcc_linux-64.sh sets LDFLAGS to have the directory
+# entries for finding libpython, liblapack, etc., so we can't
+# overwrite it
+export LDFLAGS="-shared $LDFLAGS"
+
+# conda's activate-gfortran_linux-64.sh sets F95, but
+# numpy.distutils.fcompiler looks for F90
+# this is hopefully a temporary hack
+if [ ! -v F90 ]; then
+    if [ ! -v F95 ]; then
+        echo "build.sh:error: neither F90 nor F95 variable set"
+        exit 1
+    fi
+    export F90="$F95"
+fi
+
+$PYTHON setup.py install
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -6,19 +6,16 @@ build:
   number: 1
 
 requirements:
-  build:
+  host:
     - python
     - numpy
-    - lapack
-    - m2w64-gcc-fortran # [win]
-    - gfortran_linux-64	# [linux]
-    - gfortran_osx-64	# [osx]
+    - {{ compiler('fortran') }}
+    - openblas
 
   run:
     - python
     - numpy
-    - lapack
-    - libgfortran
+    - libopenblas
 
 test:
   imports:


murrayrm:
I tried getting `conda build` to work on MacOS 10.13.5 but had problems.  When I do
```
conda build conda-recipes
```
the entire build process is carried out and it looks like everything is fine up until the final test.  At the very end I get the error
```
import: 'slycot'
Fatal Python error: PyThreadState_Get: no current thread
```
This error is occurring on the final step of the build process, where conda attempts to import slycot.  Near as I can tell from looking online this is a problem with my local python installation, but I have not yet been able to sort it out.

murrayrm:
Quick update: I am having the same problem doing a "standard" build: if I use `python setup.py build` followed by `python runtests.py` then I get the same error.  So not directly related to conda?

roryyorke:
I know nothing about OS X, but from https://github.com/snap-stanford/snap/issues/88#issuecomment-364539380 it looks like this is a Slycot build error.

I did change the build flags in build.sh (LDFLAGS and FFLAGS) - perhaps one of those is the culprit?

Separately, I'm still struggling to get Slycot to build on Windows.

repagh:
I found the import slycot error with the PyThreadState_Get issue. 
We use -lpython... to link the unknown symbols (OSX by default does not allow undefined symbols for a dynamic library, which a python module essentially is). This works if Python actually uses the libpython dylib. New conda has a python executable that has all basic python symbols statically linked in (if you are going to use them anyway, why not....). To fix the build (for now on repagh/slycot, master) I removed the library dependency, and added 
`-Wl,-dylib,-undefined,dynamic_lookup` linker flags. That solved the load problem. 

Use of openblas is interesting, where do you get it from, and is it built for all three platforms? I know I tried about a year ago, but could not get the conda openblas for windows. Maybe things are better now, since we can use compilers+linkers from conda.

roryyorke:
There is openblas for Windows, but only on conda-forge.

For Slycot, I've been trying to start with the instructions at

  https://pav.iki.fi/blog/2017-10-08/pywingfortran.html

but without much luck yet.

The `{{ compiler('fortran') }}` in meta.yaml doesn't map to
m2w64-gcc-fortran.  And one still needs MSVC.

So while this may be progress on Linux, it looks like regression on
Windows.

repagh:
On the repagh/Slycot master branch (I know, I shouldn't), I got the Windows build working. Still need to figure out how to nicely git-push that on your work. 

Solution chosen:

- compilation of the Fortran code with Flang. To get this to work, needs defines for F77 and F90 in the build script, otherwise numpy's setup script fails, since that does not "know" Flang yet. Flang can be installed through conda, and is a dependency in meta.yaml
- linking and compilation of c parts with visual studio community edition 2015. This needs a separate install of the community edition (I put the download link in the meta.yaml, without that, MicroSoft tries to push the latest and greatest on you). The associated conda package sets all environment variables correctly.
- building against openblas from conda-forge. Had to make some small tweaks in slycot/setup.py

The current conda packages are in the "repa" conda channel.

murrayrm:
It sounds like we are going to have a problem supporting different installation environments.  One tool that we might use is `python-config --ldflags`, which can be used to ask the local python installation what flags to use.  More info here: https://docs.python.org/3/extending/embedding.html#compiling-and-linking-under-unix-like-systems.

murrayrm:
This PR is superseded by PR #26.

PR #25, opened 2018-06-20
working TD04AD.f, does not crash any more on edge cases (N=0, NR=0),

repagh:
While working on supporting MIMO tf -> MIMO ss, I found some fixes had to be done to TD04AD.f

Notably edge cases (empty matrices) caused problems, 
Conversion with 'C', did not always work. 
Added a few tests in new test_td04ad.py file, checked against Octave result
corrected transform.pyf, to not allocate zero-dim matrices
matching changes in transform.py


murrayrm:
I like the idea of making it configurable, to allow maximum flexibility for people trying to install the code.  At the same time, we are going to need conda-forge versions that people can easily download (and so we have to have things working there as well).

The commit history on this is getting pretty complex, with some of the commits above seeming to related to conda/blas/lapack issues rather than the core `TD04AD.f` fix that I think was the original point of this PR.

Is it worth refactoring our pull requests a bit so that we have one that is associated with Travis CI, one associated with installation more generally (including Windows), on associated with `TD04AD.f`?

repagh:
Wow. It seems we are down to the intermittent "nontrivial imaginary part" error now, which will go away with the new common_den() function in control. 

- The Python 2.7 and 3.5 CONDA=1 versions were failing because the conda test section (import slycot) would install python 3.6 for the tests. I pinned that now.
- Since there is a slycot in conda-forge, and we need to add the conda-forge channel for lapack or openblas, there is a risk that a failing build of slycot results in installation of the conda-forge slycot, which is then being used for the further test. Re-worked the .travis.yml to fix that. 
- Lapack in conda-forge is apparently packed with blas, no need to add blas as a dependency in the conda recipe. 

I agree it has been a messy process, chasing one problem after another. When the td04ad work failed to build properly on travis, I got carried away fixing all kinds of conda and travis issues. Part of the problem was that I had no travis test environment, so it would just be down to committing and pushing. It would be cleaner to refactor, but there would be quite a bit of work involved, and it would be difficult to break things down again to steps that end up at sound endpoints, where (at least part of) the tests are passed. I am pretty confident on the current result, and I fear disentangling the work again might introduce some new errors. 

So if you don't object, I would rather have this one and the matching change to python-control merged. That should clear up the Nontrivial Imaginary Part, and lead to some progress on the other PR's too.

murrayrm:
FYI: I checked out repagh/Slycot and tested out the changes on MacOS.  I was able to build using conda, but had to explicitly set the python version:
```
```

.travis.yml
```diff
@@ -12,14 +12,17 @@ env:
 
 before_install:
   #
-  # Install fortran compiler
+  # Install fortran compiler, if not using Conda's
   #
-  - sudo apt-get install gfortran
+  #- sudo apt-get install gfortran
```
murrayrm: Suggest removing this line rather than commenting it out.

.travis.yml
```diff
@@ -50,13 +53,17 @@ install:
...
+  # lapack cannot be found in the check
+  # with --override-channels to make sure the locally built slycot is installed
   #
   - if [[ $TEST_CONDA == 1 ]]; then
-      conda install -c conda-forge lapack;
+      conda config --append channels conda-forge;
       conda build --python "$TRAVIS_PYTHON_VERSION" conda-recipe;
-      conda install slycot --use-local;
+      conda install -c conda-forge lapack;
+      conda install --override-channels -c local slycot;
```
murrayrm: Why this way of doing a local install instead of using ``--use-local`` flag?  I believe ``--use-local`` is what is in the standard conda recipe template.

murrayrm:
I got an error when trying to compile using `python setup.py install`.  Compilation was OK, but then I tried to load `slycot` (inside `ipython`), I got an error:
```
ImportError: dlopen(/Users/murray/anaconda/envs/test-repagh/lib/python3.5/site-packages/slycot/_wrapper.cpython-35m-darwin.so, 2): Library not loaded: @rpath/libopenblas.dylib
  Referenced from: /Users/murray/anaconda/envs/test-repagh/lib/python3.5/site-packages/slycot/_wrapper.cpython-35m-darwin.so
  Reason: image not found
```
Is there something that should be getting installed that isn't being installed?

repagh:
openblas from conda-forge?

.travis.yml
repagh: Yes. But --use-local *adds* the local channel to your set of channels. If the slycot build failed, it would happily install the slycot found in other channels! This happened in the python 2.7 and 3.5 variants.
Since in the overviews of installed modules I saw "local" listed as a channel for the locally built slycot, I assumed that that channel was simply called local. And it works. 

Also, --override-channels, which you need to block all others, will not work with --use-local.

murrayrm:
This PR was split into PR #25 (TD04AD) and PR #26 (conda).

PR #34, opened 2018-07-17
add support for ag08bd, sb10jd and tg01fd

marcuslil:


murrayrm:
Any possibility you could add some unit tests?

marcuslil:
Unit tests for the three functions added.

marcuslil:
I tried the max(shape(*,0),1) for ld*, but as I understand you this does not help me. Is this a critical issue? Is it solved in other slycot functions?

ilayn:
Ah sorry, let me clarify. I was just mentioning the fact that, suppose an input matrix is the empty matrix `a = np.array([], dtype=float)` for some reason. 

Then when you supply this to the low level fortran routines, because now the shape is `(0,)`, f2py will try to match `lda` to shape `(a,0)` which is `0`. In some architectures, this causes segfaults instead of Python errors. For example, if you look at the documentation of [`AG08BD`](http://slicot.org/objects/software/shared/doc/AG08BD.html) you can see that they already have `LDA = MAX(L,1)` or LDE or other `LD#` variables. That's what I meant. Hence it is a good practice to guard these variables with the following

    integer intent(hide),depend(a) :: lda=MAX(shape(a,0),1)

instead of the current `shape(a,0)`

https://github.com/marcuslil/Slycot/blob/40ccf9142711a9c85739621c85e112fd20acd711/slycot/src/analysis.pyf#L396

marcuslil:
I have corrected the MAX(ldx,1) issue. Anything else missing?

PR #38, opened 2018-08-21
Scikit buildsystem

repagh:
After many iterations, a better set of changes for using scikit-build. 
Works with conda build and with direct python/pip builds

This still contains a workaround for the lack of recent (>= 0.6.1) scikit-build on conda-forge. Directly installs & uninstall scikit-build using pip as part of a conda build. Can be simplified later.

### DIFF
--- a/conda-recipe-openblas/bld.bat
+++ b/conda-recipe-openblas/bld.bat
@@ -2,16 +2,89 @@
 cd %RECIPE_DIR%
 cd ..
 
-set F77=%BUILD_PREFIX%\Library\bin\flang.exe
-set F90=%BUILD_PREFIX%\Library\bin\flang.exe
+:: until scikit-build on conda-forge is updated to 0.6.1 or higher ...
+"%PYTHON%" -m pip install "https://github.com/scikit-build/scikit-build/archive/0.7.1.zip"
 
-"%PYTHON%" setup.py build 
+:: indicating fortran compiler is essential
+set FC=%BUILD_PREFIX%\Library\bin\flang.exe
+
+:: The batch file created by conda-build sets a load of environment variables
+:: Building worked fine without conda; apparently one or more of these 
+:: variables produce test & link failures. Resetting most of these here
+set ARCH=
+set BUILD=
+set BUILD_PREFIX=
+set CMAKE_GENERATOR=
+set CommandPromptType=
+set CPU_COUNT=
+set DISTUTILS_USE_SDK=
+set folder=
+set cpu_optimization_target=
+set fortran_compiler=
+set Framework40Version=
+set FrameworkDir=
+set FrameworkDIR64=
+set FrameworkVersion=
+set FrameworkVersion64=
+set ignore_build_only_deps=
+set CFLAGS=
+set CXXFLAGS=
+set cxx_compiler=
+set c_compiler=
+set INCLUDE=
+set LDFLAGS_SHARED=
+set LIBPATH=
+set LIB=;%LIB%
+set MSSdk=
+set MSYS2_ARG_CONV_EXCL=
+set MSYS2_ENV_CONV_EXCL=
+set NETFSXDIR=
+set PIP_IGNORE_INSTALLED=
+set platform=
+set WindowsLibPath=
+set WindowsSdkDir=
+set CYGWIN_PREFIX=
+set SRC_DIR=
+set STDLIB_DIR=
+set SUBDIR=
+set SYS_PREFIX=
+set target_platform=
+set UCRTVersion=
+set UniversalCRTSdkDir=
+set VCINSTALLDIR=
+set vc=
+set win=
+set VisualStudioVersion=
+set VSINSTALLDIR=
+set VSREGKEY=
+set VS_MAJOR=
+set VS_VERSION=
+set VS_YEAR=
+set WindowsSDKLibVersion=
+set WindowsSDKVersion=
+set WindowsSDKExecutablePath_x64=
+set WindowsSDKExecutablePath_x86=
+
+:: information on remaining variables
+set
+
+set BLAS_ROOT=%CONDA_PREFIX%
+set LAPACK_ROOT=%CONDA_PREFIX%
+
 "%PYTHON%" setup.py install
+
+:: remove scikit-build again, don't want to include that
+"%PYTHON%" -m pip uninstall --yes scikit-build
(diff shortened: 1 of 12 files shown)


moorepants:
FYI, this will hopefully be merged today: https://github.com/conda-forge/scikit-build-feedstock/pull/11

moorepants:
FYI, conda forge now has scikit-build 0.8.0.: https://github.com/conda-forge/scikit-build-feedstock/pull/14

repagh:
Yes, works. Am using the conda-forge scikit-build now, with a conda install for CONDA=0, and through the meta.yaml for CONDA=1

repagh:
@murrayrm , @moorepants, can this be merged? It uses the scikit-build currently in conda-forge (no longer needs the hack), builds and runs successfully on Linux, OSX, and Windows.

roryyorke:
I can't get this to build on Ubuntu 18.04 64-bit.  Conda version 4.5.11, conda build 3.16.0.  Build command `conda build -c conda-forge --python=3.6 slycot/conda-recipe-openblas` (from directory containg slycot).  The error I get is lengthy, but I think the important bit is
```
  File "/home/rory/.miniconda3/conda-bld/slycot_1539023903021/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/python3.6/site-packages/skbuild/command/bdist_wheel.py", line 6, in <module>
    from wheel import archive as _wheel_archive
ImportError: cannot import name 'archive'
```

I've attached a complete build log.  Any ideas what I'm doing wrong?
[build-3.6.log](https://github.com/python-control/Slycot/files/2461754/build-3.6.log)

moorepants:
> @murrayrm , @moorepants, can this be merged? It uses the scikit-build currently in conda-forge (no longer needs the hack), builds and runs successfully on Linux, OSX, and Windows.

LGTM

moorepants:
Actually I have one question. Since setup.py has changed and now requires scikit build, this needs to be made an explicit build dependency of the package. And will it work without using conda and conda-forge?

roryyorke:
And today it builds.  conda-build 3.16.1, and scikit-build 0.8.1 (previously 0.8.0).  I've got the build log if anyone's interested.

setup.py
```diff
@@ -5,7 +5,7 @@
 Slycot wraps the SLICOT library which is used for control and systems analysis.
 
 """
-from __future__ import division, print_function
+from skbuild import setup
```
moorepants: The installation documentation will need to let users know of this new requirement, as well as cmake (and any other build tools that now need to be installed). I personally think that skbuild should be an optional requirement because a user can build it without this if they so desire.

repagh: I added information on scikit-build and cmake requirements to the readme file.

I don't think it is feasible to maintain two buildsystems. I got to use cmake/skbuild after repeatedly trying to use and tweak the numpy distutils, however I found no way to specify flang to the distutils buildsystem, while flang works wonderfully for Windows builds. 

The buildsystem does not rely on conda; if skbuild and cmake are available, local builds work fine.

moorepants: Ok, seems reasonable.

roryyorke:
this had a failure on Dec. 8 https://travis-ci.org/python-control/Slycot/jobs/465308309 ; I reran one case (Python 3.6, TEST_CONDA=1) today without changing anything, and it passed https://travis-ci.org/roryyorke/Slycot/builds/473027170.  Will try a rebuild (feels a bit pointless, but what the heck...).

conda-recipe-openblas/bld.bat
```diff
@@ -2,16 +2,78 @@
...
-"%PYTHON%" setup.py build 
 "%PYTHON%" setup.py install
 
 if errorlevel 1 exit 1
 
 :: Add more build steps here, if they are necessary.
 
 :: See
-:: http://docs.continuum.io/conda/build.html
+:: http://docs.continlsuum.io/conda/build.html
```
roryyorke:
```suggestion
:: https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html
```

repagh:
There seems to be something wrong with the conda-build for python 2.7. Have not found a fix for that yet.

roryyorke:
If there are no objections to this in the next week, I propose merging to master.

In favour:
  - possible to build on Windows; hopefully we can add Windows packages to conda-forge (@moorepants, is this practical?)
  - better build output; we don't have that 1000 (or whatever) character link command appearing in the build anymore (this is fairly minor)

Two downsides:
  - scikit-built doesn't support building wheels
  - scikit-build also doesn't support "setup.py develop" mode

moorepants:
> is this practical?

I am not quite sure. As far as I know things that rely on Fortran are not possible to compile on Windows with the conda forge system. But maybe things have changed in the last months. The last I could understand was that flang might be the only way forward there.

moorepants:
It looks like the numpy feedstock installs from the numpy wheel on pypi (which is build with ATLAS on windows) and doesn't actually build on appveyor. The best bet would be to build slycot wheels for windows manually and then have the conda forge recipe install them.

ilayn:
Since v.1.0 we are able to build SciPy on Windows via OpenBLAS. It is still definitely not a walk in the park to build the BLAS of choice but if you just use a prebuilt BLAS flavor [it is quite possible](http://scipy.github.io/devdocs/building/windows.html) (I will modify the notes with a correction soon https://github.com/scipy/scipy/issues/9669).

moorepants:
It'd be great to get that working on conda forge so that other packages that need to build on windows would have a working example.

repagh:
I use a pretty bare Windows (in a virtualbox) to build the conda packages with scikit, with the following manually installed:
- Installation of msvc 2015 for the c parts
- miniconda3 
- git, but only for getting the software
For the build this uses the conda-forge flang (available win64 only), cmake, scikit-build, etc. I don't know what the conda-forge windows buildsystem looks like, but I would guess this would be compatible.

moorepants:
As soon as this is merged we can give it a shot on conda-forge and see what happens.

roryyorke:
Sorry, I thought I saw an issue about scikit-build and wheels, but the only one I can find now is closed.

I may be wrong about "setup.py develop" too; see, e.g., https://github.com/scikit-build/scikit-build/issues/371

Issue #41, opened 2018-11-18
`conda install` cannot install from local


autodrive:
Dear all,

May I have your opinion on installing Slycot using `conda` after `conda build`?

Q1: are these two lines equivalent ?
``` sh
conda install --override-channels -c local slycot
conda install --use-local slycot
```
Q2 : if Q1 is yes, and the first line above is not working for python=2.7, is a new Pull Request desirable?

If you are interested in how I became to have these questions :

I tried to fork from this repository (7f5dc1c) (again) and triggered Travis-CI.  However, the `conda install` seems unable to install from the local build; you would be able to check my test result here &rarr; https://travis-ci.org/autodrive/Slycot/jobs/456615608.

So when I manually followed the commands in .travis.yml for python=2.7 && TEST_CONDA==1.

``` sh
      conda config --append channels conda-forge;
      conda build --python "$TRAVIS_PYTHON_VERSION" conda-recipe;
      conda install -c conda-forge lapack;
      conda install --override-channels -c local slycot;
```

The result of the last line was as follows

```
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - slycot
  - libgfortran-ng[version='>=7,<8.0a0']
  - slycot
  - numpy

Current channels:

  - file:///home/jonwood/miniconda/conda-bld/linux-64
  - file:///home/jonwood/miniconda/conda-bld/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

Then I retried the following command: (Q1: is this equivalent to the original intention of 05f3ba1c ?)
``` sh
conda install --use-local slycot
```

This time the `conda` seemingly could install as follows:
``` sh
Solving environment: done

## Package Plan ##

  environment location: /home/jonwood/miniconda

  added / updated specs: 
    - slycot


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    cryptography-2.4.1         |   py27h1ba5d50_0         606 KB
    xz-5.2.4                   |       h14c3975_4         366 KB
    libxml2-2.9.8              |       h26e45fe_1         2.0 MB
    icu-58.2                   |       h9c2bf20_1        22.5 MB
    lz4-c-1.8.1.2              |       h14c3975_0         158 KB
    lzo-2.10                   |       h49e0be7_2         313 KB
    zstd-1.3.7                 |       h0b5b093_0         887 KB
[...]

PR #42, opened 2018-12-07
conda install --use-local slycot

kangwonlee:
* This PR intends to repair #41 `conda install --override-channels -c local slycot` being [unable to find the local `slycot`](https://travis-ci.org/kangwonlee/Slycot/jobs/455874419#L822) for [`python 2.7 && TEST_CONDA=1`](https://travis-ci.org/kangwonlee/Slycot/jobs/455874419).
* After this change, the build could find the local `slycot`but [one](https://travis-ci.org/python-control/Slycot/jobs/464862675#L1278) or [two](https://travis-ci.org/python-control/Slycot/jobs/464885225#L1963) of tests still may fail.
* Please see if this makes the code more harmonious.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ install:
       conda config --append channels conda-forge;
       conda build --python "$TRAVIS_PYTHON_VERSION" conda-recipe;
       conda install -c conda-forge lapack;
-      conda install --override-channels -c local slycot;
+      conda install --use-local slycot;
     else
       LAPACKLIBS=lapack:blas python setup.py install;
     fi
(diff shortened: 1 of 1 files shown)


roryyorke:
Thanks for this contribution.

I think the problem this addresses is fixed in d23ab7b, and Travis CI is now passing.  I used "local::slycot" instead of "--use-local slycot"; the "channel::package" form is not well-documented, but seems more explicit about where the package will come from.

kangwonlee:
Thanks for d23ab7b and the comment. `conda install <channel>::<package>` seems interesting, too.

PR #50, opened 2019-03-16
Extend td04ad test

repagh:
replacement for the previous PR. 
Still having a wrong commit message, as per 

https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch


repagh:
I am out of tricks, still getting the wrong commit message, with the right changes in the file though. Can we merge anyway?

PR #51, opened 2019-03-16
td04ad: testcase for issue #6

repagh:
Addition of test case, based on issue #6


roryyorke:
This looks good.  There's a case to be made for testing for correct behaviour and marking this test as should fail, but the comment is clue enough.

Issue #61, opened 2019-05-13
Install from PyPI fails - 0.3.4.0


jgspiro:
In a clean python environment, slycot does not install, because of a missing dependency (skbuild).

```
ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-er4161h2/slycot/setup.py", line 8, in <module>
        from skbuild import setup
    ImportError: No module named 'skbuild'
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-er4161h2/slycot/
```

Tried with python 2.7 and 3.5

This is not the case with previous version 0.3.3

jgspiro:
Both `scikit-build` and `numpy` need to be installed manually to be able to install slycot. ( Linux, clean python virtual env.)

ilayn:
You need to have `scikit-build` package installed.

PR #64, opened 2019-05-23
 Better handling of missing dependencies on install and README improvements.

moorepants:
- Raise an ImportError with informative message if scikit-build is not
  installed prior to executing setup.py.
- Include numpy in the install_requires of the setup function so that it
  will be installed if not present.
- Updated trove classifiers.
- Simplified and added clarity to the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,7 +16,6 @@ Slycot
 Python wrapper for selected SLICOT routines, notably including solvers for
 Riccati, Lyapunov, and Sylvester equations.
 
-
 Dependencies
 ------------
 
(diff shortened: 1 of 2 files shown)


roryyorke:
This looks good, thanks.

One point worth discussing is the switching of development status from alpha to beta.  @murrayrm, I think this is fine, even overdue---are you ok with it?

murrayrm:
Happy to upgrade the development status on PyPi to beta.  Does it appear anywhere else?   Update now or after this PR gets pushed out?

roryyorke:
I'd like to fix #61 and #3, and push out 0.3.5 ASAP - how about we make the PyPI change when that is released?

PR #66, opened 2019-06-18
Add packages entry to setup.py metadata

jakevdp:
Without this entry, pip install fails to add the package to the Python path.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -229,6 +229,7 @@ def setup_package():
 
     metadata = dict(
         name='slycot',
+        packages=['slycot', 'slycot.tests'],
         cmake_languages=('C', 'Fortran'),
         version=VERSION,
         maintainer="Slycot developers",


moorepants:
We're getting a build error on Travis:

```
$ sh -e /etc/init.d/xvfb start
sh: 0: Can't open /etc/init.d/xvfb
The command "sh -e /etc/init.d/xvfb start" exited with 127.
...
Could not connect to any X display.
The command "cd control; python setup.py test" exited with 1.
```

I think this is there to allow matplotlib to run. Maybe we can switch to setting the matplotlib to a backend that doesn't require a display?

moorepants:
This looks like a fix: https://benlimmer.com/2019/01/14/travis-ci-xvfb/

murrayrm:
The `python-control` package implements the new Travis CI format for xenial:

  https://github.com/python-control/python-control/pull/298

Should just need to replicate this in `.travis.yml`.

roryyorke:
@jakevdp thank you!  With this, a `pyproject.toml` and some CMake tweaks I can get "pip install <tarball>" to work.
I'm going to add the .travis.yml fix from python-control/python-control#298 to this PR.

jakevdp:
Maybe related (I was thinking about doing another PR) I needed to make this tweak to get ``pip install .`` to work correctly on the system I was working in:
```
$ !sed -i "s/from distutils.sysconfig import get_python_lib as pl; print(pl())/import os,numpy; print(os.path.dirname(numpy.__path__[0]))/" Slycot/CMakeLists.txt
```
This makes certain that the compilation finds the f2py sources even if numpy is not installed in the standard distutils path. I'm not certain whether it would introduce issues on other systems, though.

jakevdp:
Thanks!

PR #67, opened 2019-06-18
Build fixes: query numpy path directly; setup pyproject.toml for pip

roryyorke:
This is intended to fix #3, #63, and #65.  It works on Ubuntu 18.04, I'll still try to test it on Windows 10.  I don't have an iOS (MacOS? OSX? ... ) machine to test on.

The test is to prepare a source tarball with `python setup.py sdist` -- this needs numpy and scikit-build.  Then, in a fresh environment *without* those packages, or anything other than `pip`, run `pip install slycot.0.3.blah.tar.gz` (substitute version number for blah).  `pip install .` might be the same thing, I'm not sure.

PR #69, opened 2019-06-21
fix CMakeLists.txt to find numpy and f2py

bnavigator:
Some distributions install architecture specific python packages into directory structures like /usr/lib64/ instead of /usr/lib/. This patch allows CMake to find numpy and f2py in these directories.

CMake probably provides more sophisticated routines to find the correct python packages, but this patch is the closest to the existing routine I could come up with.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,5 +46,15 @@ if(WIN32)
   string(REPLACE "\\" "/" PYTHON_SITE ${PYTHON_SITE})
 endif()
 
+# base site dir, use python installation for location specific includes
+execute_process(
+  COMMAND "${PYTHON_EXECUTABLE}" -c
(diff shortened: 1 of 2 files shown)


roryyorke:
Thanks for the PR!  Could you take a look at #67, which I'm planning to merge soon?  @jakevdp found we needed to use the numpy "root" (see #67 diff) instead of `get_python_lib` for cases when numpy is installed in a "non-standard place"---would you mind trying that to find fortranobject.c?

bnavigator:
Yes that also works, too. See the build output at
https://build.opensuse.org/package/live_build_log/home:bnavigator/python-slycot/openSUSE_Tumbleweed/x86_64
(Although I just realized from that output that there are problems with the prefix)

I did not realize that $PYTHON_SITE is indeed only used by CMake to look for the numpy path. Disregard this PR then and merge #67.

PR #72, opened 2019-07-29
Make the .pyf signature files a dependency of the wrapper generation.

rabraker:
Currently, (as far as I can tell) you have to delete the `_skbuild` directory after editing a `.pyf` file to get `f2py` to regenerate the wrapper.  This PR adds the `.pyf` signature files as a dependency of the wrapper generation, so the builds can be incremental. That should help speed up development time when wrapping a new function (or debugging an old one).


repagh:
Looks good. 
Does not seem to be related to the OSX build failure, I have no idea what is wrong with the osx build this time....

roryyorke:
Ah, nice, thanks.

PR #73, opened 2019-08-02
addition of wrapper for mb03rd

repagh:
as requested in python-control list


roryyorke:
I haven't looked at the test cases in detail, but otherwise this looks fine to me.

We can wait a few days to see if anyone will look over the test cases, otherwise go ahead and merge.

I think I'll mark all the OSX builds as being allowed to fail; they're unfortunately not currently useful :(.  I'll open an issue for that.

bnavigator:
How about testing for jobx='N' as well? It calls a different wrapper function and has a different return structure.

repagh:
I haven't been able to locate a meaningful test case for that yet. Will try to do some more digging for that, any pointers are appreciated.

bnavigator:
Shouldn't this go into `math.py` instead of `transform.py`?

slycot/src/transform.pyf
```diff
@@ -528,3 +528,40 @@ subroutine tg01fd_uu(compq,compz,joba,l,n,m,p,a,lda,e,lde,b,ldb,c,ldc,q,ldq,z,ld
 integer required intent(in) :: ldwork
     integer intent(out) :: info
 end subroutine tg01fd_uu
+subroutine mb03rd_n(jobx,sort,n,pmax,a,lda,x,ldx,nblcks,blsize,wr,wi,tol,dwork,info) ! in MB03RD.f
```
bnavigator: It may be just my personal opinion, but I don't think it is a good idea to introduce a specific wrapper for different job options with different parameter and return signatures when there is no use case and no unit test for it. Why not just keep one wrapper  for the function, test that with the default `job='U'` and deal with 'N' as soon as it becomes relevant?

slycot/tests/test_mb03rd.py
```diff
@@ -0,0 +1,63 @@
...
+
+test1_pmax = 1e3
+test1_tol = 0.01
+class test_mb03rd(unittest.TestCase):
+    def test1(self):
+        # create schur form with scipy
+        A, X = schur(test1_A)
+        Ah, Xh = np.copy(A), np.copy(X)
+        # on this basis, get the transform
+        Ar, Xr, blks, eig = transform.mb03rd(
```
bnavigator: Assert correct `eig` as well (complex?)

repagh:
Superseded by #116

PR #78, opened 2019-08-10
remove shebang in source files without main execution path

bnavigator:
These source files do nothing when executing from command line.

The shebang is unnecessary and produces [rpmlint warnings when packaging](https://build.opensuse.org/package/show/home:bnavigator:branches:devel:languages:python:numeric/python-slycot#rpm).


roryyorke:
Cool, thanks.

PR #79, opened 2019-08-10
fix invalid variable usages from wrapper outputs

bnavigator:
Some functions tried to use variables which were never defined.


slycot/transform.py
```diff
@@ -650,7 +650,7 @@ def td04ad(rowcol,m,p,index,dcoeff,ucoeff,tol=0.0,ldwork=None):
 
     kdcoef = max(index)+1
     if rowcol == 'R':
-        porm = p
+        # porm = p
```
roryyorke: We can just as well delete these?

slycot/math.py
```diff
@@ -137,19 +138,21 @@ def mb05md(a, delta, balanc='N'):
...
-    elif out[-1] < 0:
-        error_text = "The following argument had an illegal value: "+arg_list[-out[-1]-1]
-    elif out[-1] > 0 and out[-1] <= n:
-        error_text = "Incomplete eigenvalue calculation, missing %i eigenvalues" % out[-1]
-    elif out[-1] == n+1:
+    n=min(a.shape)
+    (Ar,Vr,Yr,VALr,VALi,INFO) = _wrapper.mb05md(balanc=balanc,n=n,delta=delta,a=a)    
+    if INFO == 0:
+        VAL=VALr+1J*VALi
+        return (Ar,Vr,Yr,VAL)
```
roryyorke: The docstring headline is `Ar, Vr, Yr, VALRr, VALDr = mb05md(a, delta, balanc='N')` - would you mind checking if that's wrong (it doesn't match this return), and, if so, correct it too?

bnavigator: Yes it was wrong. Before it returned `VALr` and `VALi` as real and imaginary part of the eigenvalues. My change returns the complex VAL. (With a check added in commit af8927f if it is necessary. Returns real list otherwise.)

Docstring fixed in 961f172.

bnavigator:
Sorry, with the added unit tests and some fixes in the periphery concerning the tests directoy this turned out to be a bigger PR than initially thought. Should I go ahead and include the cleanup of the other test scripts, too or split this into a different PR?

slycot/tests/test.py
```diff
@@ -64,9 +58,5 @@ def test_td04ad_static(self):
             nr,a,b,c,d = transform.td04ad(rc,nin,nout,index,den,num)
 
 
-def suite():
```
bnavigator: This is an old artifact, right? Nothing calls this function and it would fail because of undefined `TestConvert`.

Same exists in many other test scripts here.

slycot/tests/test.py
```diff
@@ -64,9 +58,5 @@ def test_td04ad_static(self):
...
-   return unittest.TestLoader().loadTestsFromTestCase(TestConvert)
```
bnavigator: This is an old artifact, right? Nothing calls this function and it would fail because of undefined `TestConvert`.

Same exists in many other test scripts here.

slycot/math.py
```diff
@@ -62,16 +63,16 @@ def mc01td(dico,dp,p):
...
+        warnings.warn('P(x) may have zeros very close to stability boundary.')
     if out[-2] > 0:
-        warnings.warn('The degree of P(x) has been reduced to %i' %(dp-k))
+        warnings.warn('The degree of P(x) has been reduced to %i' %(dp-out[-2]))
     return out[:-2]
 
 
 def mb05md(a, delta, balanc='N'):
-    """Ar, Vr, Yr, VALRr, VALDr = mb05md(a, delta, balanc='N')
+    """Ar, Vr, Yr, VAL = mb05md(a, delta, balanc='N')
```
roryyorke: Thanks for updating this.

This changes the API --- but, I'm not sure anyone is using it, in particular, python-control isn't using mb05md that I could see.

@repagh , it looks like you're the one who added this in 7b7eaad2 - do you know of users of this function?

slycot/tests/test_mb.py
```diff
@@ -0,0 +1,82 @@
...
+        (Ar, Vr, Yr, VAL) = mb05md(A, delta)
+
+        assert_allclose(Ar, Ar_ref, atol=0.0001)
+
+        # Order of eigenvalues is not guaranteed, so we check them one by one.
+        for i, e in enumerate(VAL):
+            erow = np.ones(VAL.shape)*e
+            i_ref = np.isclose(erow, VAL_ref)
+            self.assertTrue(any(i_ref),
+                            msg="eigenvalue {} not expetced".format(e))
```
roryyorke: expected, not expetced.

slycot/tests/test.py
roryyorke: I see I introduced that in 9a098d12.  I don't remember why anymore, but I see it's been cargo-culted everywhere else.  From  file:///usr/share/doc/python3/html/library/unittest.html#organizing-test-code it looks like this `suite` function *could* introduce problems, so yes, let's erase them.  Thanks!

slycot/math.py
repagh: I have forgotten what I added it for....
Typically, this is used in calculating a discrete state-space from a continuous one. I see it is currently not used in python-control. Good idea to return the eigenvalue as a complex vector.

PR #80, opened 2019-09-09
build & test using GitHub Workflows

kangwonlee:
Dear all,

Just in case, I prepared a configuration to build and test using the GitHub Workflows.

My test results would be available [here](https://github.com/kangwonlee/Slycot/runs/215921746).

Hope this helps and please let me know any suggestions and/or questions.

KW


kangwonlee:
Consider using an upper level `env`

> Env at the workflow and job level
>
>It is common to need define a set of environment variables that are used in multiple steps in a job and even multiple jobs in a workflow. Now you can add an `env` map at both the workflow and job level. Those environment variables will be merged with the `env` defined at any step lower in the hierarchy

https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features/#env-at-the-workflow-and-job-level

bnavigator:
Superseded by #140

Issue #86, opened 2020-01-20
Wrapper for functions related to periodic Schur decomposition


RRiva:
Dear Slycot support,

SLICOT is probably the only library with an implementation of the periodic Schur decomposition, which is very useful in the analysis of Linear Time-Periodic systems. The main function is `MB03WD`, which needs `MB03VD` and probably also `MB03XP` and `MB03YD`. Could you please add a wrapper for these subroutines?

Thank you very much 😊 

Best regards,
Riccardo Riva

bnavigator:
Is #88 what you are looking for? `MB03WX` is used in the documentation example, but the unittest I built based on it succeeds without it.

`MB03XP` and it's auxiliary routines could be added, too. Can't promise if I find time in the next few days though.

RRiva:
Hi Ben, yes it is! It's nice to see how short the test becomes when it's written in python. Also declaring `W` as a complex array is a good idea.

I only needed `MB03WD`, so you can skip the other routines.

Thanks a lot 😊

roryyorke:
Fixed in #88.

PR #88, opened 2020-01-26
add schur decomposition wrappers

bnavigator:
As requested by #86 

Please review. Especially the variable names of input and output matrices and the index ranges mentioned in the docstrings.


slycot/math.py
```diff
@@ -70,6 +70,361 @@ def mc01td(dico,dp,p):
...
+
+    n : int
+            The order of the square matrices A_1, A_2, ..., A_p.
+            n >= 0.
+
+    ilo, ihi : int
+            It is assumed that all matrices A_j, j = 2, ..., p, are
+            already upper triangular in rows and columns [:ilo] and
+            [ihi:n], and A_1 is upper Hessenberg in rows and columns
+            [:ilo] and [ihi:n], with A_1[ilo-1,ilo] = 0 (unless
```
roryyorke, line 97: This paragraph is fairly tough going.  Is the idea that the upper and lower blocks `A[:ilo,:ilo]` and `A[ihi:n, ihi:n]` are diagonal (or Hessenberg) already?

Is `A_1[ilo-1, ilo] = 0` part of the Hessenberg condition (in which case, shouldn't it be `A_1[ilo, ilo-2] = 0` ?), or is this an extra condition?

slycot/math.py
```diff
@@ -70,6 +70,361 @@ def mc01td(dico,dp,p):
...
+
+    A : ndarray
+            A[:n,:n,:p] must contain the matrices of factors to be reduced;
+            specifically, A[:,:,j-1] must contain A_j, j = 1, ..., p.
+
+
+    Returns
+    -------
+
+    HQ : ndarray
```
roryyorke, line 111: I think specifying the exact size of `HQ` would be good (I gather `HQ.shape` will be `(n, n, p)` ?

Similarly, could you say what the size of `Tau` will be?  (this one I couldn't guess)

slycot/math.py
```diff
@@ -70,6 +70,361 @@ def mc01td(dico,dp,p):
     return out[:-2]
 
 
+def mb03vd(n, ilo, ihi, A):
```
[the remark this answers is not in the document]
bnavigator: The idea is to leave it open and let the Fortran routine populate the `[:n,:n]` part just as it is in the docstring. The routines in analysis.py do it the same way. Infering from the size of input `A` is also possible but then restrains n>=1 instead of n>=0 like in the Fortran routine. I don't think that that would change anything substantial.

slycot/math.py
```diff
@@ -70,6 +70,361 @@ def mc01td(dico,dp,p):
...
+    """ HQ, Tau = mb03vd(n, ilo, ihi, A)
```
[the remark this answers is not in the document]
bnavigator: Yes. The Fortran routines sometimes use the same generic name for input and output where the actual formulas use different symbols.

slycot/math.py
bnavigator: This comes from the fact that Python is 0-indexed while the Fortran SLICOT routines count from 1. My request for review is to check if I made the conversion correctly.

slycot/math.py
bnavigator: `HQ.shape` is the same shape as input A. Minimum is `(max(n, 1), max(n, 1), p)` but could be larger.
`Tau.shape` is `(max(1, n-1), p)`

slycot/math.py
roryyorke: the comment header from `MB03VD.f` is:

```
C             already upper triangular in rows and columns 1:ILO-1 and
C             IHI+1:N, and A_1 is upper Hessenberg in rows and columns
C             1:ILO-1 and IHI+1:N, with A_1(ILO,ILO-1) = 0 (unless
C             ILO = 1), and A_1(IHI+1,IHI) = 0 (unless IHI = N).
C             If this is not the case, ILO and IHI should be set to 1
C             and N, respectively.
C             1 <= ILO <= max(1,N); min(ILO,N) <= IHI <= N.
```

If this is Matlab-type indexing, which I'd guess it is (otherwise I think `IHI:N` would read `IHI:N+1`), then it's inclusive of end-points (closed), while Python is half-open.  For `ILO=5`, Matlab 1:ILO-1 means [1,2,3,4]; Python [:ilo] is [0,1,2,3,4], while we want [0,1,2,3].  So should that be "upper diagonal (or Hessenberg) in `[:ilo-1]`" ? This also makes sense for the lower bound of 1 on ilo; for ilo = 1 the first `[:ilo-1] = [:0] #=> empty` rows are upper diagonal (or Hessenberg).

I think the `ihi` expressions are correct: suppose N=10, IHI=7.  Fortran: `IHI+1:N = 8:10 = [8,9,10]`; Python: `[ihi:n] = 7,8,9`.

I still think the `A[ilo-1, ilo] = 0` comment is wrong, though I see it is straight from the SLICOT comment.  I'll add another comment later on this.

slycot/math.py
roryyorke: Isn't it worth adding this to the docstring?

If `n < min(A.shape[:2])`, `HQ` would still have the same size as A?  Would the `n:` rows and columns of HQ then be uninitialized?

bnavigator: Yes that's the way many of the SLICOT routines do it. Adding this to the docstring and rebasing once more.

We have a bunch of different docstring styles for the various functions. Opening another issue for that.

slycot/math.py
roryyorke: N = 5; ilo = 2

A(ILO, ILO-1) = A(2,1) = 0

  1 2 3 4 5
1 d u u u u
2 x d u u u
3 x 0 d u u
4 x x x d u
5 x x x x d

I got this confused -- I thought the specified zero was above the diagonal.  It looks like this is an additional restriction on  A_1, i.e., beyond being only upper Hessenberg.

The Fortranesque `A_1(ILO,ILO-1) = 0` and `A_1(IHI+1,IHI) = 0` map correctly to the Pythonesque `A_1[ilo-1,ilo] = 0` and  `A_1[ihi,ihi-1] = 0`.

slycot/math.py
roryyorke: Thank you.  Do you agree with this, from https://github.com/python-control/Slycot/pull/88#discussion_r407025183:

     So should that be "upper diagonal (or Hessenberg) in [:ilo-1]" ?

slycot/math.py
bnavigator: So you now agree that `A[ilo-1, ilo] = 0` is correct?

roryyorke: Yes, that seems right.

bnavigator: Let's leave MATLAB out of here and focus on Fortran vs Python ;)

Fortran `(1,ILO-1)` has ILO-1 entries, so you're right it should be Python `[:ilo-1]`. Still confused about the `:n` now. Need another minute to think about it.

bnavigator: A[ilo-1, ilo] = 0 is NOT correct. Shouldn't it be A[ilo-1:ilo-2]?

slycot/math.py
bnavigator: Agree. General rule is:
- `Python-index = Fortran-index - 1`
- `Python-slice-start = Fortran-slice-start-1 
- `Python-slice-end:Fortran-slice-end`

Or put differently:
` A(x,y) = A[x-1,y-1]`
` A(a:b,c:d) = A[a-1:b,c-1:d]`

bnavigator: Fix committed

slycot/math.py
roryyorke: The original Fortran comment in MB03VD.f line 51 is `with A_1(ILO,ILO-1) = 0`.  I think this specifies a single entry at (1-based) row `ILO-1` and column `ILO`, not a range (`,`, not ':').  This mean `A[ilo-1, ilo] = 0` is OK?

I think this diff is still needed for `mb03vd`

```
diff --git a/slycot/math.py b/slycot/math.py
index 39a7ea7..41b8fa1 100644
--- a/slycot/math.py
+++ b/slycot/math.py
@@ -92,9 +92,9 @@ def mb03vd(n, ilo, ihi, A):
 
     ilo, ihi : int
             It is assumed that all matrices A_j, j = 2, ..., p, are
-            already upper triangular in rows and columns [:ilo] and
+            already upper triangular in rows and columns [:ilo-1] and
             [ihi:n], and A_1 is upper Hessenberg in rows and columns
-            [:ilo-1] and [ihi:n], with A_1[ilo-1,ilo-2] = 0 (unless
+            [:ilo-1] and [ihi:n], with A_1[ilo,ilo-1] = 0 (unless
             ilo = 1), and A_1[ihi,ihi-1] = 0 (unless ihi = n).
             If this is not the case, ilo and ihi should be set to 1
             and n, respectively.
        1 <= ilo <= max(1,n); min(ilo,n) <= ihi <= n.
```

bnavigator: Yes it is a single entry, not a slice. Thus, 
`FortranEntry(A_1(ILO, ILO-1)) == PythonEntry(A_1(ilo-1, ilo-2))`

On [:ilo-1], i agree. Commited.

roryyorke:
Please let me know if 1a5d38f looks right to you; it's the same indexing change we discussed for `mb03vd` applied to `mb03wd`.

bnavigator:
It looks right, but I have committed another one, also same issue as above with `[ilo-1, ilo-2]` ;)

roryyorke:
I missed that.  Ah, I now see what your earlier comment about [ilo-1, ilo-2] was, sorry.  So mb03wd also needs a fix.

Here's how things stand at 4d3a03d; I'm going to change mb03wd to match mb03vd.   

* mb03vd
for: A_1(ILO,ILO-1) = 0
py: A_1[ilo-1,ilo-2] = 0

* mb03wd
for: H_1(ILO,ILO-1) = 0
py: H_1[ilo-1,ilo]

slycot/math.py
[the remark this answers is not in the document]
roryyorke: HQ and Tau are probably fine, and the caller is anyway not obliged to use the same names as in the docstrings.

slycot/math.py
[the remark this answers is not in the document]
roryyorke: The names are fine; the caller can use his or her own names if they object.

PR #96, opened 2020-04-01
add ab08nz function to support complex state-space systems

lytex:
add ab08nz function to support regular pencil for complex state-space systems

### DIFF
--- a/slycot/analysis.py
+++ b/slycot/analysis.py
@@ -575,7 +575,7 @@ def ab08nz(n, m, p, A, B, C, D, equil='N', tol=0., lzwork=None):
         e.info = info
         raise e
     return (nu, rank, dinfz, nkror, nkrol, infz, kronr, kronl, Af, Bf,
-            int(zwork[0]))
+            int(zwork[0].real))
 
 def ab09ad(dico,job,equil,n,m,p,A,B,C,nr=None,tol=0,ldwork=None):
     """ nr,Ar,Br,Cr,hsv = ab09ad(dico,job,equil,n,m,p,A,B,C,[nr,tol,ldwork])
--- a/slycot/analysis.py
+++ b/slycot/analysis.py
@@ -475,6 +475,94 @@ def ab08nd(n,m,p,A,B,C,D,equil='N',tol=0,ldwork=None):
         raise e
     return out[:-1]
 
+def ab08nz(n,m,p,A,B,C,D,equil='N',tol=0,ldwork=None):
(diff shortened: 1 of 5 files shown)


bnavigator:
Please add one or more unit tests here as discussed in https://github.com/python-control/python-control/pull/376

lytex:
I've written a simple test for both `ab08nd` and `ab08nz`. 

`ab08nd` works fine but `ab08nz` SEGFAULTs

bnavigator:
I sent you a PR https://github.com/lytex/Slycot/pull/1. If you accept this PR here will be updated. See also https://github.com/bnavigator/Slycot/tree/complex-pyf for Travis CI results.

slycot/analysis.py
```diff
@@ -475,6 +475,108 @@ def ab08nd(n,m,p,A,B,C,D,equil='N',tol=0,ldwork=None):
...
+    right and left Kronecker indices of the system (A,B,C,D).
+
+    Required arguments:
+        n : input int
+            The number of state variables.  n >= 0.
+        m : input int
+            The number of system inputs.  m >= 0.
+        p : input int
+            The number of system outputs.  p >= 0.
+        A : input rank-2 array('d') with bounds (n,n)
```
roryyorke, line 494: "with bounds (n,n) is" not terminology familiar to me.  Maybe "with A.shape >= (n,n)" ?

slycot/analysis.py
```diff
@@ -475,6 +475,108 @@ def ab08nd(n,m,p,A,B,C,D,equil='N',tol=0,ldwork=None):
...
+        equil := 'N' input string(len=1)
+            Specifies whether the user wishes to balance the compound matrix
+            as follows:
+            = 'S':  Perform balancing (scaling);
+            = 'N':  Do not perform balancing.
+        tol := 0.0 input float
+            A tolerance used in rank decisions to determine the effective rank,
+            which is defined as the order of the largest leading (or trailing)
+            triangular submatrix in the QR (or RQ) factorization with column
+            (or row) pivoting whose estimated condition number is less than 1/tol.
```
roryyorke, line 516: Maybe add something about the lower bound on tol, say "if tol is too small (including the default value of 0), a tolerance based on machine precision and problem size is used" ? Otherwise the combination of default `tol = 0` and `1/tol` is a bit confusing, and I think misleading.

slycot/analysis.py
bnavigator: See #100. It comes directly from f2py and is used in many other existing docstrings in that file.

slycot/analysis.py
bnavigator: I re-added the paragraph from the Fortran comment in https://github.com/bnavigator/Slycot/tree/pr96-lytex-complex

Issue #44, opened 2019-01-18
Fix Coveralls URL in README.md


roryyorke:
The coveralls link is broken; it's not just the case of "Slycot" in the URL, since both https://coveralls.io/r/python-control/slycot (the current URL in README.md) and https://coveralls.io/r/python-control/Slycot give 404s.

I don't know much about coveralls; is python-control/Slycot registered properly?  Are we missing a step in the Travis CI build?

We could just remove this badge and URL.

kangwonlee:
* If I may; briefly checked and noticed its (sort-of) [upper level link](https://coveralls.io/github/python-control/) lists [python-control](https://github.com/python-control/python-control) only.
* IMHO, one of https://github.com/orgs/python-control/people could take a look at the [upper level link](https://coveralls.io/github/python-control/)?
* Please let me know any possible improvements

murrayrm:
Good catch @kangwonlee.  I've added `slycot` to the list of repositories in the `python-control` account on Coveralls.  As soon as there is another build, that should fix things up.

kangwonlee:
* I could see that https://coveralls.io/github/python-control/Slycot is now available
* Any further action necessary?

roryyorke:
Thanks @kangwonlee , I'd forgotten this.  Something is still not quite right, since coveralls says "no data" for all those builds.

We probably need to tweak the Travis build to find the results of the "Slycot/runtests.py" command.

roryyorke:
Fixed in #97 and #101.

PR #97, opened 2020-04-03
switch the additional python-control unittests to pytest in Travis

bnavigator:
python-control switched its unit test runs to pytest. Let's do that in Slycot's CI run as well.

I could not test out how this affects coveralls. Depending on the outcome 7a1d846 might need to be reverted.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,12 @@
 [run]
 source = slycot
-omit = */tests/*
+omit = 
+  */tests/*
+  */version.py
+
+
+# please do not add any sections after this block
+# the CI will add the slycot modules as last line here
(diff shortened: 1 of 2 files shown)


bnavigator:
coveralls does still not combine the filepaths from different environments correctly. Will try to work on it later.

bnavigator:
With all the rebasing and reverting, the link to travis seems to be lost. Here it is:

https://travis-ci.org/github/python-control/Slycot/builds/671057317

I think this is ready to be merged.

bnavigator:
This should also fix #44 as soon as it is merged and a CI build of master is reported to coveralls

PR #93, opened 2020-02-04
reorganize travis configuration, remove runtests.py

bnavigator:
Linux builds:
- conda builds
- based on pip and Ubuntu 18.04 packages
- Python versions 2.7 to 3.8

OSX builds:
- CONDA with custom SDK (curretly not functional, see discussion on #76 and PR #92)
- native osx / pip packages

Replaced old runtests.py with direct pytest call

(conflicts with #92)

PR #91, opened 2020-02-01
Work on build system

bnavigator:
Reopen of #90 rebased onto master.

This still needs some tweaking with the .travis.yml file.

PR #101, opened 2020-04-12
Update README.rst build instructions

roryyorke:
- note tests depend on scipy
- remove reference to deleted conda-build
- added conda instructions per platform (Linux, macOS, Windows)
- remove instruction installing "plain" LAPACK from conda
- minor rewording and fixes


roryyorke:
@bnavigator please review; this should be as discussed in #98

README.rst
```diff
@@ -87,34 +89,36 @@ before running the install::
...
 
-If you are using conda, you can also get working (binary) copies of LAPACK from
-conda-forge using the command::
+    conda build conda-recipe-openblas
+    conda install --use-local slycot
 
-   conda install -c conda-forge lapack
+Additional tips for how to install Slycot from source can be found in the
+``.travis.yml`` (commands used for Travis CI) and conda-recipe/ (conda
+pre-requisites) both which are included in the source code repository.
```
bnavigator, line 117:
```suggestion
Additional tips for how to install Slycot from source can be found in the
``.travis.yml`` (commands used for Travis CI) and the ``conda-recipe-*/``
directories (conda pre-requisites) both which are included in the source
code repository.
```

README.rst
```diff
@@ -87,34 +89,36 @@ before running the install::
...
     python setup.py install
 
-You can also use conda to build and install Slycot from source::
+You can also use conda to build and install Slycot from source, but
+you'll have to choose the right recipe directory.
 
-    conda build conda-recipe
-    conda install --use-local slycot
+On Linux you can choose between conda-recipe-openblas and
+conda-recipe-mkl
```
bnavigator, line 101:
```suggestion
On Linux you can choose between ``conda-recipe-openblas`` and
``conda-recipe-mkl``
```

README.rst
```diff
@@ -87,34 +89,36 @@ before running the install::
...
 
-If you prefer to use the OpenBLAS library, a conda recipe is available in
-``conda-recipe-openblas``.
+On macOS you should use conda-recipe-apple.
```
bnavigator, line 103:
```suggestion
On macOS you should use ``conda-recipe-apple``.
```

README.rst
```diff
@@ -87,34 +89,36 @@ before running the install::
...
 
-Additional tips for how to install Slycot from source can be found in the
-``.travis.yml`` (commands used for Travis CI) and conda-recipe/ (conda
-pre-requisites) both which are included in the source code repository.
+On Windows you can try either conda-recipe-openblas or
+conda-recipe-mkl; they allow for Windows builds, but we don't
```
bnavigator, line 106:
```suggestion
On Windows you can try either ``conda-recipe-openblas`` or
``conda-recipe-mkl``; they allow for Windows builds, but we don't
```

bnavigator:
The coveralls badge also needs updating (fixes #44): 

```
.. image:: https://coveralls.io/repos/github/python-control/Slycot/badge.svg
   :target: https://coveralls.io/github/python-control/Slycot
```

README.rst
bnavigator: Actually, I have no idea, whether the windows builds work at all. The `bld.bat` in `conda-recipe-openblas` is unchanged from when it was still ``conda-recipe`` and uses environment variables for Generic BLAS/LAPACK. conda-forge/slycot-feedstock also has the same setup.

@moorepants, @murrayrm, @repagh you are maintaining the feedstock. Care to comment?

Should we move the Windows Script to `conda-recipe-win/`? Anybody got Slycot with MKL (with conda provided libs?) set up in Windows?

roryyorke:
@bnavigator thank you.

bnavigator:
Quoting my own comments from the review above, so they don't get buried.

> On line 80 it should read "Unpack the **s**ource code ..."

> Actually, I have no idea, whether the windows builds work at all. The `bld.bat` in `conda-recipe-openblas` is unchanged from when it was still `conda-recipe` and uses environment variables for Generic BLAS/LAPACK. conda-forge/slycot-feedstock also has the same setup.
> 
> @moorepants, @murrayrm, @repagh you are maintaining the feedstock. Care to comment?
> 
> Should we move the Windows Script to `conda-recipe-win/`? Anybody got Slycot with MKL (with conda provided libs?) set up in Windows?

bnavigator:
Discussion in #98 going on: We need to clarify the description how to build on MacOS X

roryyorke:
conda-build-openblas doesn't work on my Windows system.  First I hit this [1] (which I think we've discussed before); I added `build-backend="setuptools.build_meta" ` to pyproject.toml, but then CMake couldn't find Visual Studio 2015, though it *could* find it when I built in a conda environment with `python setup.py install` (see #102).

I'm inclined to explicitly state we don't have a Windows conda build recipe; we can point to the conda-forge Slycot feedstock for anyone who wants to try to build their own conda package.

I also don't want to give detailed Windows build instructions; the script I have is hacky, and I suspect fragile.

I won't be able to give much time to this until Saturday.  Meanwhile anyone wanting to wrap this PR up and make the 0.4.0 release is welcome to.

[1] https://github.com/pypa/setuptools/issues/1694

bnavigator:
> I'm inclined to explicitly state we don't have a Windows conda build recipe; we can point to the conda-forge Slycot feedstock for anyone who wants to try to build their own conda package.

+1 for that. That means we should also remove the `bld.bat` in both conda recipe directories.

roryyorke:
@bnavigator Is pytest (now?) a pre-requisite for running the unit tests?  I saw you used a `@pytest` decorator in #104.

bnavigator:
Yes, should probably add it to the various dependency declarations.

bnavigator:
https://github.com/python-control/Slycot/blob/003a2afe09987d80e45302bd3534eb530653f542/conda-recipe-apple/meta.yaml#L10

Since you're already cleaning up the conda recipes in this PR, the `_mkl_` infix needs to be changed to `_apple_`

roryyorke:
I haven't done anything about https://github.com/python-control/Slycot/issues/98#issuecomment-612641670:

> #91, #92, #93 also have some back and forth discussion about getting Mac to build. Turns out it is crucial, that the correct MacOSX SDK is used. Maybe we should also add this to the README.rst in #101.

I'll skim the referenced issues / PRs, but otherwise suggestions for wording for this are weclome.

README.rst
```diff
@@ -87,35 +77,74 @@ before running the install::
...
 
-If you prefer to use the OpenBLAS library, a conda recipe is available in
-``conda-recipe-openblas``.
+On Linux you can choose between ``conda-recipe-openblas`` and
+``conda-recipe-mkl``
 
-Additional tips for how to install Slycot from source can be found in the
-``.travis.yml`` (commands used for Travis CI) and conda-recipe/ (conda
-pre-requisites) both which are included in the source code repository.
+On macOS you should use ``conda-recipe-apple``.
```
bnavigator, line 94:
```suggestion
On macOS you should use ``conda-recipe-apple``. See the the `conda-build documentation`_ how to get the required macOS SDK.

.. _conda-build documentation: https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#macos-sdk>
```

README.rst
```diff
@@ -87,35 +77,74 @@ before running the install::
...
+    conda create --channel conda-forge --name build-slycot-py35 python=3.5 scikit-build flang numpy scipy
+    activate build-slycot-py35
+
+    REM you won't need this if conda installs f2py.exe rather than f2py.cmd or f2py.bat
+    where f2py > %TMP%\F2PYPATH.txt
+    set /P F2PYPATH=< %TMP%\F2PYPATH.txt
+
+    python setup.py install -- -DF2PY_EXECUTABLE=%F2PYPATH%
+    cd slycot\tests
+    python -m unittest -v
```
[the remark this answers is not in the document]
roryyorke: OK - presumably one also then install pytest at environment creation time.

README.rst
```diff
@@ -87,35 +77,74 @@ before running the install::
...
+dependencies, *except* for the C compiler.
 
-If you are using conda, you can also get working (binary) copies of LAPACK from
-conda-forge using the command::
+First, install the appropriate Visual Studio compiler (see
+https://wiki.python.org/moin/WindowsCompilers ).
 
-   conda install -c conda-forge lapack
+In the unpacked Slycot source directory, run the following commands.
+This example is for Python 3.5; adapt as required.
```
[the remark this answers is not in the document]
roryyorke: This is what I tested with.  I need to test the latest master (and the pytest mod) anyway, so maybe I'll bite the bullet and install the newest VS build tools.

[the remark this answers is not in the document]
bnavigator: Nothing for this release, but would it be worth to switch from Travis to Github Workflows and include Windows builds there? Possibly based on #80, although to me that seems a bit outdated by now.

README.rst
```diff
@@ -87,35 +77,77 @@ before running the install::
...
+    cd slycot\tests
+    python -m unittest -v
+
+General notes on compiling
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Additional tips for how to install Slycot from source can be found in the
+``.travis.yml`` (commands used for Travis CI) and the ``conda-recipe-*/``
+directories (conda pre-requisites) both which are included in the source
+code repository.
```
bnavigator, line 134: Is that true? Not sure here, but wouldn't you need to list them in MANIFEST.in

README.rst
```diff
@@ -32,48 +32,38 @@ following dependencies:
 
 - Python 2.7, 3.5+
 - NumPy
-- scikit-build >=0.8.1
-- cmake
+- scikit-build
```
moorepants, line 35: I noticed this: https://github.com/conda-forge/slycot-feedstock/blob/master/recipe/meta.yaml#L32, so you may need to leave a version specifier.

README.rst
```diff
@@ -32,48 +32,38 @@ following dependencies:
...
+binaries from conda-forge (see below); these are available for Linux,
+OSX, and Windows.
 
-Install Miniconda or Anaconda and then Slycot can be installed via the conda
-package manager from the conda-forge channel with the following command::
-
-    conda install -c conda-forge slycot
-
-Using pip
-~~~~~~~~~
```
moorepants, line 62: Seems you removed the note about pip, is it no longer supported?

README.rst
```diff
@@ -87,35 +77,77 @@ before running the install::
...
 
-    conda build conda-recipe
-    conda install --use-local slycot
+You can also use conda to build and install Slycot from source, but
+you'll have to choose the right recipe directory.
 
-If you prefer to use the OpenBLAS library, a conda recipe is available in
-``conda-recipe-openblas``.
+On Linux you can choose between ``conda-recipe-openblas`` and
+``conda-recipe-mkl``
```
moorepants, line 92: Note that there seems to be relatively new way for users to switch between blas versions: https://conda-forge.org/docs/maintainer/knowledge_base.html#blas

README.rst
```diff
@@ -87,35 +77,77 @@ before running the install::
...
+Using pip
+~~~~~~~~~
+
+We publish Slycot to the Python package index, but only as a source
+package, so to install using pip you'll first need to install the
+build prerequisites (compilers, libraries, etc.)
+
+If you have these build prerequisites, install in the standard way with:
+
+    pip install slycot
```
moorepants, line 153: Scratch my pip comment above, I see now it was moved.

README.rst
roryyorke: Thanks, will do.

README.rst
bnavigator: That's only for users getting their packages from conda-forge not for using conda-build locally, though? So rather something for the conda-feedstock package than for the instructions here?

moorepants: Sure, but it begs the question for me why Slycot wants to try to provide conda build instructions for non-defaults/conda-forge dependencies. Building a compatible set of conda binaries locally has to pull dependencies from somewhere or you have to build the whole stack locally. Mixing the underlying blas/lapack/etc libs with those in conda-forge and defaults is likely asking for trouble.

What is the reason for having these conda build scripts in slycot at this point in time? Doesn't the setup.py that's now powered by scikit build let users do custom installations sufficiently?

bnavigator: Totally agree, that the current mixing is asking for trouble.

The linked article implies that it should not matter against which library the build is linked during compile time, because the implementations should be ABI compatible. Maybe we are overcomplicating things here and if we still want to provide a recipe just give a single default one?

roryyorke: I'm not sure why we maintain these conda-build recipes, especially given the existence of slycot-feedstock.

moorepants: I think a single recipe is possible. It is done on the conda forge feedstock, for example. If the user is willing to use conda-forge to pull dependencies, theoretically they can conda build the recipe in the feedstock.

moorepants: > I'm not sure why we maintain these conda-build recipes, especially given the existence of slycot-feedstock.

It is legacy code, from before conda forge in some sense. I used to have conda build recipes in all of my software projects, but have now abandoned it if favor of the conda forge feedstocks.

bnavigator: +1 for not providing recipes directly.

roryyorke: > If the user is willing to use conda-forge to pull dependencies, theoretically they can conda build the recipe in the feedstock.

Is it reasonably easy to use the recipe in the feedstock to do test builds of master (and other branches)?  My understanding was that these recipes were quite strict about what they built: don't they verify  a checksum of the source, or something similar?

If it *is* possible to use the feedstock recipe, we should base our CI and build instructions on it.

moorepants: > Is it reasonably easy to use the recipe in the feedstock to do test builds of master (and other branches)?

The recipes in the feedstock are normal conda recipes. I just tested on Ubuntu 20.04. I cloned the feedstock recipe, cd'd into the recipe directory, and did `conda build -c conda-forge .`. It built successfully and passed the tests. There are some parts of the recipe that are specific to conda forge, e.g. disabling the python 2.7 builds for windows, which a user would have to remove if they wanted to do that build.

repagh:
I think it would be best to refer to the windows build recipe on slycot-feedstock, and not have those in the slycot source code tree. It would only make sense if we have slycot developers who use windows for development. For me, it was and is a PITA to create the windows recipe, I will hack on the feedstock one for my students & the general public, but have no interest in maintaining a recipe in the slycot tree as well.

murrayrm:
I agree with the comment that we don't need to spend a lot of time on the Windows build if (a) we have condo-forge working and (b) nobody is developing on Windows.  We can just refer interested parties to the slycot-feedstock if they want help.

I do think that we should keep the MacOS version working (see #98), but here also I would focus on what active developers are using (eg, we can delete conda-recipe-apple unless an active developer is using it).  For me (not currently active, but sometimes a contributor), it is more useful to have the setup.py script working for MacOS versus a conda build script.

roryyorke:
Is there consensus to remove *all* the conda recipes, and add a pointer in README.rst to slycot-feedstock for anyone needing such a recipe?

Along with this we would also change the Travis config to only use "setup.py" type builds.

README.rst
roryyorke: > The recipes in the feedstock are normal conda recipes. I just tested on Ubuntu 20.04. I cloned the feedstock recipe, cd'd into the recipe directory, and did conda build -c conda-forge .. It built successfully and passed the tests. There are some parts of the recipe that are specific to conda forge, e.g. disabling the python 2.7 builds for windows, which a user would have to remove if they wanted to do that build.

OK, but at the top of the recipe is this:

```
{% set version = "0.3.5.0" %}
{% set sha256 = "cad98d5ea4f0a034cf398c39189f587620a0b03f1d4b71e77cd622a327f13adf" %}

package:
  name: slycot
  version: {{ version }}

source:
  url: https://pypi.io/packages/source/s/slycot/slycot-{{ version }}.tar.gz
```

which means 0."build 3.5 from PyPI", right? This is appropriate for the feedstock, but doesn't easily let one build other versions.

This is all OK, I just want to be sure I understand.

README.rst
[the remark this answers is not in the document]
roryyorke: Latest (cb791af513e4d38bd950b849acfb2a34e9a432b2) instructions specify pytest

README.rst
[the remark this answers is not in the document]
roryyorke: Examples changed to Python 3.8 (also tested on 3.7).  We've got #108 for using Github Workflows.

README.rst
roryyorke: As of cb791af513e4d38bd950b849acfb2a34e9a432b2 this is specified.

README.rst
```diff
@@ -32,48 +32,38 @@ following dependencies:
...
 
-    pip install slycot
+    conda install -c conda-forge slycot
 
-From source
-~~~~~~~~~~~
+From source without conda (Linux, macOS, Windows)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Unpack the course code to a directory of your choice,
```
bnavigator, line 68:
```suggestion
Unpack the source code to a directory of your choice,
```

roryyorke: Hah! Thanks.

README.rst
moorepants: A user can change the version number and see if it builds, but if you want to make old builds you would likely need to checkout a prior commit in the feedstock recipe corresponding to a previous version. But, that said, keeping old builds working as time passes will get increasingly hard due to updates in conda, conda build, and any dependencies. You'd have to build with old versions of conda and put max versions on the dependencies, etc.

bnavigator: I think the question is more about building a more recent version from the git repository instead of the released 0.3.5 (or hopefully soon 0.4.0). Even then it is just a matter of adjusting the source tag.

moorepants: Yes, the general practice in conda forge is to point ot "official" releases on PyPi (so that it parallels what `pip install` does).

bnavigator: Of course, because that is what you want to have in the conda channel.

Now, if a user wants to compile a newer version of Slycot, say because there is a new feature in master or even in a PR that has not been released yet. Can he or she take the conda-forge recipe, use it for the checked out commit in question and install it in his or her environment? Sure. Question is, should we provide instructions for it or?

repagh:
I like to keep the conda recipes in. These can be a reference for the feedstock ones, and since conda-forge is an important distribution channel for slycot, I would prefer to have a conda build in the tests.

PR #109, opened 2020-04-26
Fix unit tests option 1: move tests directory and do not install them.

bnavigator:
Option 1 as discussed in https://github.com/python-control/Slycot/issues/98#issuecomment-619635645

Do not install the tests module. Only run `pytest` after install.


roryyorke:
Went with #110.

PR #110, opened 2020-04-26
Fix unit tests option 2: run pytest with --pyargs slycot by default

bnavigator:
Option 2 as discussed in https://github.com/python-control/Slycot/issues/98#issuecomment-619635645

Make sure `pytest` tests with the installed module found in `sys.path`, not the local source dir without the compiled wrapper.


roryyorke:
See also https://github.com/python-control/Slycot/issues/98#issuecomment-619638141

Issue #100, opened 2020-04-11
Docstring standards in Slycot


bnavigator:
We have a bunch of different docstring styles used in the various definitions.

Some directly come from the `__doc__` property of the wrapper functions generated by f2py, recognizable by the `bounds` qualifier.
https://github.com/python-control/Slycot/blob/196e4ec6a5b77e612f8d2ee7082fe1e27a0dfb25/slycot/math.py#L104

Other functions use `shape()`
https://github.com/python-control/Slycot/blob/196e4ec6a5b77e612f8d2ee7082fe1e27a0dfb25/slycot/synthesis.py#L74

Sometimes we have blank space between the parameter definitions. Sometimes not.

What's your opinion on unifying this? Adhere to [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)  and get closer to the documentation in numpy and scipy? Decide on a proper way to specify the shape? 

The array('d') is also not consistent with numpydoc, I guess.

Would also affect the math formulas.

moorepants:
+1 to adopting numpydoc.

bnavigator:
I used numpydoc in #103

bnavigator:
With the recent PRs of @KybernetikJo, this has come a long way. Thank you!

PR #117, opened 2020-05-03
Add Slycot exception and warning classes

### LINKED ISSUE #99: Replace ValueError with info attribute with SlycotError
See https://github.com/python-control/Slycot/pull/84#issuecomment-611991035

A common pattern in the Slycot code is

```
        e = ValueError('the number of eigenvalues to be assigned is less than the number of possibly assignable eigenvalues')
        e.info = info
        raise e
```

The idea is to have a SlycotError class, so that this can be:

```
    raise SlycotError(msg, info)
```
instead - that way one can't forget the .info attribute.

I'd have had `SlycotError` inherit from `RuntimeError`, but for backward compatibility it should probably be `ValueError`.

bnavigator:
As proposed in #99, I am introducing the new Exceptions `SlycotError`, `SlycotParameterError` and `SlycotArithmeticError`. So to never forget the `.info` attribute again!

Fixes #99


slycot/analysis.py
```diff
@@ -121,28 +124,21 @@ def ab01nd(n,m,A,B,jobz='N',tol=0,ldwork=None):
     if jobz == 'N':
         out = _wrapper.ab01nd_n(n,m,A,B,tol=tol,ldwork=ldwork)
         if out[-1] < 0:
-            error_text = "The following argument had an illegal value: "+arg_list[-out[-1]-1]
-            raise ValueError(error_text)
+            raise SlycotParameterError(None, out[-1], arg_list)
```
repagh, line 127: Looks a bit ugly and hackish

Would it be an idea to use named parameters?
```
raise SlycotParameterError(info=out[-1], arglist=arg_list)
```
Or call 
```
SlycotParameterError(arg_list, out[-1])
```
then detect that the first parameter is not a string and compose a message?

slycot/analysis.py
```diff
@@ -679,11 +660,11 @@ def ab09ad(dico,job,equil,n,m,p,A,B,C,nr=None,tol=0,ldwork=None):
     else:
         ordsel = 'F'
     if dico != 'C' and dico != 'D':
-        raise ValueError('Parameter dico had an illegal value')
+        raise SlycotParameterError('Parameter dico had an illegal value', -1)
```
repagh, line 663: Here you are essentially inventing new info numbers. What will they be used for?

repagh: Ah no, I see, they correspond to the SLICOT parameter number. OK, ignore my remark.

slycot/analysis.py
bnavigator: Thanks for looking at this!

I wanted to make it consistent with the `("msg", info)` order of parameters. 

I personally don't like the overloading of a single constructor but that would be possible.

Another option: Have an additional subclass derived from `SlycotParameterError` with `SlycotGenericParameterError(arg_list, info)`. Maybe there is a better idea for the name, though.

Or get rid of the ever same
```python

if info < 0:
    raise SlycotParameterError(....)
```

and replace it with a call of

```python

def check_illegal_parameter(arg_list, info):
    if info < 0:
         message = ("The following argument had an illegal value: {}"
                    "".format(arg_list[-info-1]))
         raise SlycotParameterError(message, info)
```

repagh: That last solution is the way to go!

repagh: I am getting another idea, can we format the messages from the "Raises" section in the functions docstring? That would force good docstrings, and eliminate the typing!

```python
def verify_slycot_return(arg_list, docstring, info):
    if info < 0:
         message = ("The following argument had an illegal value: {}"
                    "".format(arg_list[-info-1]))
         raise SlycotParameterError(message, info)
    elif info > 0:
         # process the docstring for the error message
         # ....
```

repagh: p.s. I am working on this now...

repagh:
Ow. That went quicker than I thought. It seems I was able to push on your branch, if hope you like the modification, if not, ignore/revert. 

One of the advantages is that this reduces the number of lines; fewer lines - better coverage.

bnavigator:
I like it, but right now the raise from docstring always raises `SlycotParameterError` even when it should be `SlycotArithmeticError`. Working on this right now

repagh:
I think I accidentally did a wrong force-push, trying to fix the loose ampersand. :-(

bnavigator:
Yeah, you discarded my edits. No worries, will fix that in my next commit. Probably need to rebase a little bit.

bnavigator:
> Look, nice improvement:
> 
> ```
> coverage/coveralls — Coverage increased (+25.1%) to 68.396%
> ```

It's cheating though because we now parse stuff that is not counted by coverage. ;)

repagh:
Nice work, you lifted it to a whole new level.

Some additional points
- I see you removed the parsing check from test_sb.py. I worked on an easy routine to verify that a docstring can be parsed for the exceptions that we think are in there. That would offset the "cheating"? 
- I am getting these complaints from git that the recent commit messages have a & and a <, fix or ignore?

bnavigator:
Yes the original parsing check did not work anymore as the new parser does not return the complete list.

Will fix the commit messages in the next rebase, where I also fix your "merge" commit. Right now I am getting gihub connection errors on git.

I am working on extending this to "Warns" also. There are quite a few routines that use the info parameter for warnings but want to return partial results.

slycot/exceptions.py
```diff
@@ -0,0 +1,228 @@
...
+    SlycotArithmeticError:
+    Info is 2, which is between 1 and 4
+    >>> fun(4)
+    SlycotArithmeticError:
+    4 is in [4,    1.2e+02)!
+    >>> fun(120)
+    SlycotResultWarning:
+    120 is too large
+    """
+    if docstring:
```
repagh, line 212: shouldn't we check info > 0?
Now we are parsing the docstring after each calculation.

slycot/exceptions.py
```diff
@@ -0,0 +1,228 @@
...
+        exception, message = _parse_docsection("Raises", docstring, checkvars)
+        if exception and message:
+            fmessage = '\n' + message.format(**checkvars).strip()
+            raise globals()[exception](fmessage, info)
+
+        warning, message = _parse_docsection("Warns", docstring, checkvars)
+        if warning and message:
+            fmessage = '\n' + message.format(**checkvars).strip()
+            warn(globals()[warning](fmessage, info))
+
```
repagh, line 224: and if we don't find a warning or message, and info > 0, we should return a catch-all exception, with just the info number. Better safe than sorry.

slycot/math.py
```diff
@@ -584,12 +556,9 @@ def mb03wd(job, compz, n, ilo, ihi, iloz, ihiz, H, Q, ldwork=None):
...
-    if info < 0:
-        e = ValueError(
-                "Argument '{}' had an illegal value".format(arg_list[-info-1]))
-        e.info = info
-        raise e
-    elif info > 0:
+    raise_if_slycot_error(info, arg_list)
+
+    if info > 0:
         warnings.warn(("failed to compute all the eigenvalues {ilo} to {ihi} "
```
repagh, line 562: can also be replaced?

slycot/math.py
```diff
@@ -609,63 +578,78 @@ def mb05md(a, delta, balanc='N'):
...
+        appear consecutively with the eigenvalue having positive
+        imaginary part first.
+
+    Warns
+    ------
+    SlycotResultWarning : e
+        :0 < e.info <=n:
+            the QR algorithm failed to compute all
+            the eigenvalues; no eigenvectors have been computed;
+            w[{e.info:n}] contains eigenvalues which have converged;
```
repagh, line 646: [{e.info}:{n}]

slycot/exceptions.py
bnavigator: Custom messages with info < 0 should be possible

slycot/math.py
bnavigator: Yes, there are a lot of functions which can now be simplified. I am currently going through 
`analysis.py`.

I wanted to do the docstring cleanup (#100) in a separate PR later, but now with all the modifications in the docstrings already, which is not that easy to review it is a good thing to do much work on that now.

slycot/exceptions.py
bnavigator: I agree, that now with the possibility to mention warnings, a catchall can be implemented.

slycot/math.py
```diff
@@ -18,6 +18,8 @@
 #       MA 02110-1301, USA.
 
 from . import _wrapper
+from .exceptions import raise_if_slycot_error, SlycotArithmeticError
```
repagh, line 21: SlycotArithmeticError no longer needed

slycot/math.py
```diff
@@ -552,22 +520,22 @@ def mb03wd(job, compz, n, ilo, ihi, iloz, ihiz, H, Q, ldwork=None):
...
-    ValueError : e
-            e.info contains information about the exact type of exception
-
+    Warns
+    -----
+    SlycotResultWarning
+        :info > 0:
+            failed to compute all the eigenvalues {ilo} to {ihi}
+            in a total of 30*({ihi}-{ilo}+1) iterations
+            the elements Wr{{info}:{ihi}] contains those
```
repagh, line 537: Wr{{info}:{ihi}]  to Wr[{info}:{ihi}]

slycot/exceptions.py
repagh: ok, but I am worried about performance
can we do
```python
    if docstring and (iwarn or info):
        ....
```
then we don't parse the docstring upon succesful calculation. Might matter to someone who is calculating a lot of LQR controllers or the like.

slycot/synthesis.py
```diff
@@ -286,20 +279,25 @@ def sb02md(n,A,G,Q,dico,hinv='D',uplo='U',scal='N',sort='S',ldwork=None):
...
+        :info = 4:
+            The Hamiltonian or symplectic matrix H has less
+            than n stable eigenvalues;
+        :info = 5:
+            The n-th order system of linear algebraic
+            equations, from which the solution matrix X would
+            be obtained, is singular to working precision.
 
 
         Example
```
repagh, line 303: detent 4 spaces

slycot/math.py
```diff
@@ -487,32 +485,25 @@ def mb03wd(job, compz, n, ilo, ihi, iloz, ihiz, H, Q, ldwork=None):
...
             H_j, j = 2, ..., p.
-
-    Q : ndarray
+    Q : array_like
             If compz = 'V', Q[:n,:n,:p] must contain the current matrix Q of
             transformations accumulated by SLICOT Library routine
             MB03VY.
             If compz = 'I', Q is ignored
-
     ldwork : int, optinal
```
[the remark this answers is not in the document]
bnavigator:
```suggestion
    ldwork : int, optional
```

slycot/math.py
bnavigator:
```suggestion
from .exceptions import raise_if_slycot_error
```

slycot/math.py
bnavigator:
```suggestion
            the elements Wr[{info}:{ihi}] contains those
```

repagh:
Looks perfect now. Any objections to a merge?

bnavigator:
As soon as you said it's perfect, I broke it ;) Fixed now.

With all the numpydoc changes, this is hard to review. Let's give @roryyorke a little bit of time to see if he has any objections.

slycot/math.py
```diff
@@ -681,9 +680,9 @@ def mb05nd(a, delta, tol=1e-7):
 
     Returns
     -------
-    F : ndarray
+    F : (n n) ndarray
```
repagh, line 683: ,

bnavigator:
```suggestion
    F : (n, n) ndarray
```

bnavigator:
So, time to merge and focus on the release before we get new ideas? ;)

PR #121, opened 2020-05-11
Update coveralls badge

bnavigator:
The coveralls badge still seems to have problems.


repagh:
Matches the doc I see on coveralls.io. Thanks.

PR #127, opened 2020-05-17
[WIP] override XERBLA to raise exception

### LINKED ISSUE #126: SB10FD exits whole process on parameter error
In reference to https://github.com/python-control/Slycot/issues/98#issuecomment-629778408:

```
[ben@voyagerS9:~]% python                                                                                                   [0]
Python 3.8.2 (default, Apr  8 2020, 14:31:25) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from slycot import synthesis
>>> from numpy import array
>>> A = array([[-1, 0, 0], [0, -12, -5], [0, 4, 0]])
>>> B = array([[2, 0], [0, 0.5], [0, 0]])
>>> C = array([[-0.5, 0, -0.5], [0, 0, 0], [-0.5, 0, -0.5]])
>>> D = array([[0, 0],  [0, 1], [0, 0]], dtype=float)
>>> synthesis.sb10fd(3, 2, 3, 1, 1, 100, A, B, C, D, 1e-7, None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ben/src/Slycot/_skbuild/linux-x86_64-3.8/cmake-install/slycot/synthesis.py", line 2717, in sb10fd
    raise_if_slycot_error(out[-1], arg_list, sb10fd.__doc__)
  File "/home/ben/src/Slycot/_skbuild/linux-x86_64-3.8/cmake-install/slycot/exceptions.py", line 236, in raise_if_slycot_error
    raise globals()[exception](fmessage, info)
slycot.exceptions.SlycotArithmeticError: 
The Y-Riccati equation was not solved
successfully (the controller is not admissible or
there are numerical difficulties)
>>> synthesis.sb10fd(1,1,1,0,1,1,[1],[1],[1],[1],1e-3,1)
 ** On entry to SB10FD parameter number 20 had an illegal value
[ben@voyagerS9:~]%                                                                                                          [0]
```

The last drop out of the whole Python process should not happen. A simple Exception should be enough. Probably have to patch the Fortran code?

bnavigator:
Fixes #126 (eventually)

Trying to wrap the Fortran `XERBLA` call. Still having some issues with the parameter list.

```
Python 3.8.2 (default, Apr  8 2020, 14:31:25) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import slycot
>>> print(slycot._wrapper.xerbla.__doc__)
xerbla(srname,info,raise_xerbla,[raise_xerbla_extra_args])

Wrapper for ``xerbla``.

Parameters
----------
srname : input string(len=-1)
info : input int
raise_xerbla : call-back function

Other Parameters
----------------
raise_xerbla_extra_args : input tuple, optional
    Default: ()

Notes
-----
Call-back functions::

  def raise_xerbla(srname,info): return 
  Required arguments:
    srname : input string(len=-1)
    info : input int

>>> def raise_xerbla_2_arguments(srname, info):
...     raise Exception("If I was called with ({}, {}) you would see this".format(srname, info))
... 
>>> slycot._wrapper.raise_xerbla = raise_xerbla_2_arguments
>>> slycot.sb10fd(1,1,1,0,1,1,[1],[1],[1],[1],1e-3,1)
capi_return is NULL
Call-back cb_raise_xerbla_in___user__routines failed.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ben/src/Slycot/_skbuild/linux-x86_64-3.8/cmake-install/slycot/synthesis.py", line 2714, in sb10fd
    out = _wrapper.sb10fd(n, m, np, ncon, nmeas, gamma,
TypeError: raise_xerbla_2_arguments() missing 2 required positional arguments: 'srname' and 'info'

```

But:
```
Python 3.8.2 (default, Apr  8 2020, 14:31:25) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import slycot
>>> def raise_no_arguments():
...     raise Exception("no arguments provided")
... 
>>> slycot._wrapper.raise_xerbla = raise_no_arguments
>>> slycot.sb10fd(1,1,1,0,1,1,[1],[1],[1],[1],1e-3,1)
capi_return is NULL
Call-back cb_raise_xerbla_in___user__routines failed.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ben/src/Slycot/_skbuild/linux-x86_64-3.8/cmake-install/slycot/synthesis.py", line 2714, in sb10fd
    out = _wrapper.sb10fd(n, m, np, ncon, nmeas, gamma,
  File "<stdin>", line 2, in raise_no_arguments
Exception: no arguments provided
>>> 
```

### DIFF
--- a/slycot/CMakeLists.txt
+++ b/slycot/CMakeLists.txt
@@ -104,7 +104,7 @@ set(FSOURCES
     src/delctg.f  src/select.f
     src/SLCT_DLATZM.f  src/SLCT_ZLATZM.f
 
-    src/ftruefalse.f
+    src/ftruefalse.f src/XERBLA.f
 )
 
 set(F2PYSOURCE src/_wrapper.pyf)
--- a/slycot/__init__.py
+++ b/slycot/__init__.py
@@ -43,6 +43,11 @@ else:
     # Version information
     from .version import version as __version__
 
+    # initialize error handling
+    from .exceptions import raise_xerbla
+    from . import _wrapper
+    _wrapper.raise_xerbla = raise_xerbla
+
     from numpy.testing import Tester
     test = Tester().test
     bench = Tester().bench
--- a/slycot/exceptions.py
+++ b/slycot/exceptions.py
@@ -257,3 +257,11 @@ def raise_if_slycot_error(info, arg_list=None, docstring=None, checkvars=None):
         warn(SlycotWarning("Caught unhandled nonzero IWARN value {}"
                            "".format(iwarn),
                            iwarn, info))
+
+
+def raise_xerbla(srname, info):
+    """Overrides LAPACK XERBLA routine to raise Exception instead of exiting
+    """
+    message = ("The argument number {1} to {0} had an illegal value."
+               "".format(srname, -info))
+    raise SlycotParameterError(message, info)
--- a/slycot/src/XERBLA.f
+++ b/slycot/src/XERBLA.f
@@ -1,0 +1,19 @@
+      SUBROUTINE XERBLA(SRNAME, INFO)
+C
+C     SLYCOT
+C     Override LAPACK XERBLA routine to raise Python Exception instead of
+C     exiting the process
+C
+CF2PY INTENT(CALLBACK, HIDE) RAISE_XERBLA
+C
+      CHARACTER*(*) SRNAME
+      INTEGER INFO
+C
+      EXTERNAL RAISE_XERBLA
+C
+      CALL RAISE_XERBLA(SRNAME, INFO)
+C
+      RETURN
+C
+      END
+C
--- a/slycot/src/_helper.pyf
+++ b/slycot/src/_helper.pyf
@@ -6,5 +6,10 @@ subroutine ftruefalse(ftrue,ffalse) ! in src/ftruefalse.f
     logical intent(out) :: ffalse
 end subroutine ftruefalse
 
-! This file was auto-generated with f2py (version:2).
-! See http://cens.ioc.ee/projects/f2py2e/
+subroutine xerbla(srname,info) ! in src/XERBLA.f
+    use __user__routines
+    character*(*) :: srname
+    integer :: info
+    intent(callback) raise_xerbla
+    external raise_xerbla
+end subroutine xerbla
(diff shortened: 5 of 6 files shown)


bnavigator:
https://github.com/numpy/numpy/issues/16277

repagh:
OK, is this anyhow possible? How does the exception propagate back through the Fortran calls? AFAIK Fortran has no exception handling.
Then how are Python exceptions handled, and would this not mess with the call stack?

bnavigator:
The idea is to use [F2PY callback arguments](https://numpy.org/devdocs/f2py/python-usage.html#call-back-arguments). The Fortran routine `XERBLA` calls `RAISE_XERBLA` which is supposed to be the Python function `slycot._wrapper.raise_xerbla` which can then raise the Exception.

The simpler solution would be to just make `XERBLA` noop to avoid the annoying print to stdout and exit. We can always let the python code handle the INFO parameter as we already do.

repagh:
I don't have the insight in Python's exception mechanism to determine if it is even possible and safe to pass an exception with Fortran routines on the F2PY call stack. 

If you have the experience for that, by all means, but I would take the safe route.

bnavigator:
> If you have the experience for that,

I don't ;)

Yo have a point here:

I tested by executing
```python
>>> import slycot
>>> import numpy as np
>>> n=1000; m=3; p=2
>>> A = np.random.randn(n,n); B = np.random.randn(n,m)
>>> C = np.random.randn(p,n); D=np.random.randn(p,m)
>>> for i in range(500):
...     try:
...         slycot.sb10fd(n,m,p,0,1,1,A,B,C,D,1e-3,1)
...     except TypeError:
...         pass
... 
```
For this test, it doesn't matter if the exception currently raised is the desired SlycotError or the TypeError from my initial post above. This allocates a lot of memory during the loop and never frees it until the python process is ended.

So you are probably right: Raising exceptions in a callback and thus never returning to the Fortran call stack is possible, but maybe not a good idea.

I wonder if there will be any comments by F2PY people, reacting to my question in numpy/numpy#16277

PR #129, opened 2020-05-19
Fix ab01nd

bnavigator:
This function fails for `jobz='N'`, because you cannot assign to tuple members.

First a test function to demonstrate...

### DIFF
--- a/slycot/tests/test_ab08n.py
+++ b/slycot/tests/test_ab08n.py
@@ -1,5 +1,5 @@
 # ===================================================
-# ab08nX tests
+# ab08n* tests
 
 import unittest
 from slycot import analysis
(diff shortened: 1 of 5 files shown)


bnavigator:
... now with the fixed and cleaned function.

I converted the docstring to numpydoc and removed the duplicate wrapper code for different `jobz` values.

slycot/tests/test_ab08n.py
```diff
@@ -1,5 +1,5 @@
 # ===================================================
-# ag08bd tests
+# ab08nX tests
```
repagh, line 2: There is also an AB08NX.f in the SLICOT sources. Should we then later have an ab08nX and an ab08nx test file?

bnavigator:
```suggestion
# ab08n* tests
```

PR #130, opened 2020-05-20
expand test_xerbla_override to allow debugging

bnavigator:
Be more verbose if the test fails. It could fail because of xerbla not being overridden or because the import of the wrapper library failed.

### DIFF
--- a/slycot/tests/test_exceptions.py
+++ b/slycot/tests/test_exceptions.py
@@ -24,6 +24,7 @@ import sys
 
 from slycot.exceptions import raise_if_slycot_error, \
                               SlycotError, SlycotWarning, SlycotParameterError
+from slycot import _wrapper
 
 
 def assert_docstring_parse(docstring, exception_class, erange, checkvars={}):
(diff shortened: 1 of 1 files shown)


repagh:
Yes, it now works fine. I only get errors when running from within the Slycot folder; maybe that was the problem all along. 
```
Slycot repa$ pytest --pyargs slycot
```
From the result:
```
input = None, capture_output = False, timeout = None, check = True
popenargs = (['/Users/repa/anaconda3/envs/test-slycot/bin/python', '-c', '\nfrom slycot._wrapper import ab08nd\n# equil=\'X\' is invalid\nout = ab08nd(1, 1, 1, [1], [1], [1], [1], equil=\'X\')\nprint("INFO={}".format(out[-1]))\n'],)
kwargs = {'stderr': -2, 'stdout': -1, 'universal_newlines': True}, process = <subprocess.Popen object at 0x1177f0ad0>
stdout = 'Traceback (most recent call last):\n  File "<string>", line 2, in <module>\n  File "/Users/repa/control/src/Slycot/sl...er\nImportError: cannot import name \'_wrapper\' from \'slycot\' (/Users/repa/control/src/Slycot/slycot/__init__.py)\n'
stderr = None, retcode = 1
```
The other tests all work, so this seems to be due to the fact that the subprocess code is executed in the current folder, where it finds a slycot sub-folder, complete with __init__.py but without wrapper.

bnavigator:
Should be able to run from source folder again

repagh:
Perfect, good solution

PR #131, opened 2020-05-20
Add a pytest header module

bnavigator:
In order to determine fast wether the correct package and wrapper library are tested, display them in the pytest header.

Needs to be installed into the distribtution as as pytest module via setuptools, because we usually run pytest from outside of the source dir.


bnavigator:
... will also display the Slycot specific stuff for all non-related pytest calls in other test suites. So not a great idea how it is implemented.

PR #133, opened 2020-05-24
Fix warnings capture

bnavigator:
Addresses the failures in https://github.com/python-control/Slycot/issues/98#issuecomment-633175028


roryyorke:
Thank you!

Issue #108, opened 2020-04-25
Consider switching CI to Github Workflows, include Windows


roryyorke:
Nothing for this release, but would it be worth to switch from Travis to Github Workflows and include Windows builds there? Possibly based on #80, although to me that seems a bit outdated by now.

_Originally posted by @bnavigator in https://github.com/python-control/Slycot/pull/101_

bnavigator:
This will also be a good opportunity to adopt [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), which effectively means to drop the support for Python 2 and 3.5 now and 3.6 very soon.

bnavigator:
I have been trying to prepare something in #140, but it is not finished yet.

Dropping Python 2 will also help to avoid workarounds like in #139.

PR #136, opened 2020-05-30
switch old numpy nosetester to pytest call

### LINKED ISSUE #134: slycot.test() does not work anymore
Ref. https://github.com/conda-forge/slycot-feedstock/pull/26#discussion_r432888422

The conda-feedstock recipe called `python -c "from slycot import test; test()"` which uses the numpy Tester, which does use nose and cannot work with the new pytest parametrizations.

bnavigator:
Fixes #134 

Numpy did something similar in https://github.com/numpy/numpy/pull/10827

I guess those lines for `.test` and `.bench` in Slycot came from and old documentation for writing numpy modules. Slycot currently does not have any benchmarks and according to the numpy PR, `bench` is not used anymore.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -136,9 +136,16 @@ To test if the installation was successful, you can run the slycot unit tests::
 
     pytest --pyargs slycot
 
-Running ``pytest`` without ``--pyargs slycot`` from inside the source directory
-will fail, unless either ``setup.cfg`` or the compiled wrapper library have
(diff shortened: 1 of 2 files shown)


roryyorke:
Changes look good to me.

I'm going to add an issue to include a test for `import slycot; slycot.test()`, but let's wait for discussion and decision on #108 before doing anything.

Issue #137, opened 2020-05-31
Add CI test for `import slycot; slycot.test`


roryyorke:
Follow-up to #136.  Let's decide on #108 before doing anything for this.

PR #140, opened 2020-12-30
GitHub workflow

bnavigator:
See https://github.com/bnavigator/Slycot/tree/github-workflow for the GitHub Actions results.

Tasks:
 - Ubuntu
   - [x] setup.py install
   - [x] Conda
   - [x] Pip
 - macOS
   - [x] Conda
   - [x] Pip
 - Windows
   - [x] Conda
 - Python Versions: 3.6 to 3.9
 
- [x] Streamline LAPACK linking. The conda-forge packages nowadays always link against the reference implementation but you can use a different one (=the same as numpy) during runtime.
- [x] do #137 : Test slycot.test()


conda-recipe-generic/build.sh
```diff
@@ -0,0 +1,3 @@
+$PYTHON setup.py build_ext install -- \
+        -DNumPy_INCLUDE_DIR=${SP_DIR}/numpy/core/include \
```
roryyorke, line 2: I just filed https://github.com/scikit-build/scikit-build/issues/524 because I think scikit-build should find the venv numpy before system numpy.

roryyorke:
Thanks for working on this!

It looks like this PR will drop Python <= 3.5 (I think Numpy is about to go >= 3.7 anyway); I think that's worth explicitly noting.

conda-recipe-generic/build.sh
bnavigator: This recipe is taken from the conda-forge feedstock.

bnavigator:
> It looks like this PR will drop Python <= 3.5 (I think Numpy is about to go >= 3.7 anyway); I think that's worth explicitly noting.

See also https://github.com/python-control/python-control/pull/481

bnavigator:
> - Streamline conda recipes / sync with conda-forge feedstock?
> - Streamline LAPACK linking? The conda-forge packages nowadays always link against the reference implementation but you can use a different one (=the same as numpy) during runtime.

@repagh, I remember that prior to the 0.4.0 release you advocated for keeping these different conda recipes. Is this still true? Do we need to test them all?

conda-recipe-generic/build.sh
bnavigator: Are we sure the problem is not the fiddling with the python path in  https://github.com/python-control/Slycot/blob/055e9d62c28801467abf896e307671ea71b78088/CMakeLists.txt#L21-L32 ? 

`find_package(NumPy )` is not a scikit-build feature but from CMake.

bnavigator: > find_package(NumPy ) is not a scikit-build feature but from CMake.

I am wrong here. Scikit-build's FindNumpy is a different thing than [FindPython with NumPy component from CMake](https://cmake.org/cmake/help/latest/module/FindPython.html).

roryyorke: Cool - that `find_package` finds the right thing without extra magic.  I've noted that on the scikit-build issue.

bnavigator: It might not help us (see my comment linking to the disussion there), because it sets different output variables in cmake. We need to find the correct f2py too. I was experimenting with FindF2PY from scikit-build yesterday. Curious if all the old windows hacks in setup.py and the CMake files are still necessary. They do not seem to work for my setup in the workflow so far.

roryyorke:
This looks good, but I'm in no position to review the details.  One thing I did wonder about: why are the test matrices generated by scripts, and not known upfront?

Will we remove Travis CI config after this is merged?

For anyone else reviewing, you can see output here: https://github.com/bnavigator/Slycot/actions

The warnings there seem to be due to a third-party package; @bnavigator has already queried it https://github.com/conda-incubator/setup-miniconda/issues/132#issuecomment-756387028

Given that free Travis CI is disappearing, I think we should merge this.  @murrayrm , @repagh, objections?

bnavigator:
> why are the test matrices generated by scripts, and not known upfront?

The test matrix generation is based upon which wheel and conda packages are created. With all the exceptions and special cases for combinations of OS and LAPACK implementation, statically defining the matrices would be very complicated.

murrayrm:
No objections to merging on my end.  I note that the full CI build seems to take ~30 minutes, but of course it is testing on all sorts of different configurations and OS's.

bnavigator:
Most of the time is wasted by the conda resolver. Especially on Windows. This could be fixed be a frozen environment, which we update manually, but that is maybe something for another PR.

PR #142, opened 2021-01-08
Clean build setup

bnavigator:
It turns out that some of the code in setup.py and the CMake files is no longer needed. Not even for Windows. (#140 is not ready yet, but it builds on this one and I already *had* successful Windows builds which also passed the unit tests.)

- Reverts a change in which I broke a MacOSX Travis run, because pyenv on Travis is too old.
- Ignores that the conda package built against mkl on Python 3.9 produces a crash. Compiling against specific LAPACK/BLAS packages is not the way to compile conda packages anymore. https://conda-forge.org/docs/maintainer/knowledge_base.html#blas
- Related: https://github.com/python-control/Slycot/pull/140#discussion_r550238923

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,46 +7,21 @@ if (CMAKE_VERSION VERSION_GREATER "3.11.99")
 endif()
 
 project(slycot VERSION ${SLYCOT_VERSION} LANGUAGES NONE)
-# Fortran detection fails on windows, use the CMAKE_C_SIMULATE flag to
-# force success
-if(WIN32)
-  set(CMAKE_Fortran_SIMULATE_VERSION 19.0)
-endif()
 
-# this does not seem to work, maybe scikit-build's doing? the cxx compiler is
-# still tested
 enable_language(C)
 enable_language(Fortran)
 
-# base site dir, use python installation for location specific includes
-execute_process(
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -7,46 +7,21 @@ if (CMAKE_VERSION VERSION_GREATER "3.11.99")
...
 find_package(PythonLibs REQUIRED)
+find_package(PythonExtensions REQUIRED)
 find_package(NumPy REQUIRED)
+find_package(F2PY REQUIRED)
 find_package(BLAS REQUIRED)
 find_package(LAPACK REQUIRED)
 
-message(STATUS "NumPy: ${NumPy_INCLUDE_DIR}")
+message(STATUS "NumPy included from: ${NumPy_INCLUDE_DIR}")
+message(STATUS "F2PY included from: ${F2PY_INCLUDE_DIR}")
```
roryyorke, line 22: `F2PY_INCLUDE_DIR` isn't documented as being set (see https://github.com/scikit-build/scikit-build/blob/c0741f9a1de9dd97e0a1e6f0b3b02fcd76221418/skbuild/resources/cmake/FindF2PY.cmake).  This is possibly just an oversight in the docs.

roryyorke (approved):
Thanks @bnavigator , looks good.

PR #147, opened 2021-02-03
Stop reporting Travis coverage results to Coveralls.

bnavigator:
The coverage from the Travis runs conflicts with the coverage from the GitHub Actions runs. Disable to report it.

Next up would be to disable Travis CI completely.


roryyorke:
LGTM.  I don't have an opinion on when to disable Travis; feel free to do so.

PR #149, opened 2021-02-07
install conda-verify

bnavigator:
Windows builds hang for several minutes, and after the hang they report, that they could not find conda-verify.

Let's see if this helps.


bnavigator:
Nope does not help with windows build time. Let's install conda-verify anyway.

(The test errors are still because of python-control/python-control#533)

PR #159, opened 2021-10-31
limit sniff-build to Python 3.9

bnavigator:
GHA defaults to Python 3.10 already, but scipy did not yet publish a wheel for the cp310 ABI.

This forces the first check of our CI to build its own scipy wheel, which takes a long time.


roryyorke:
LGTM.

I quickly looked at the coverage decrease, but couldn't see where it was; it could only be Python code that is version dependent, and I don't know if we have all much of that.  I think the decrease doesn't matter too much.

If you're also comfortable with the apparent coverage change, please merge.

PR #160, opened 2021-11-09
sunset py 3.6, test 3.10

bnavigator:
Python 3.6 has been removed from the NumPy support matrix quite some time ago ([NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)), Main Python 3.6 [reaches EOL](https://www.python.org/dev/peps/pep-0494/) next month. No reason to further torture the CI with it.

Meanwhile, SciPy wheels for Python 3.10 are available.

There are also no severe differences between Python 3.7 and 3.8. Testing the older one should be enough.

### DIFF
--- a/.github/workflows/slycot-build-and-test.yml
+++ b/.github/workflows/slycot-build-and-test.yml
@@ -55,10 +55,9 @@ jobs:
               - 'ubuntu'
               - 'macos'
             python:
-              - '3.6'
               - '3.7'
-              - '3.8'
               - '3.9'
+              - '3.10'
             bla_vendor: [ 'unset' ]
             include:
               - os: 'ubuntu'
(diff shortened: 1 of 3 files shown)


murrayrm:
This looks good to go, but we should perhaps add an issue to add back in conda testing of 3.10 when it becomes available.

bnavigator:
> add back in conda testing of 3.10 when it becomes available.

Python 3.10 is already prepared for conda-forge. As soon as https://github.com/conda-forge/slycot-feedstock/pull/39 is merged, https://github.com/conda-forge/slycot-feedstock/pull/37, can be renewed and merged too.

bnavigator:
> This looks good to go, but we should perhaps add an issue to add back in conda testing of 3.10 when it becomes available.

Done in #161

PR #168, opened 2022-01-19
Addition of Binder link

jonititan:
added binder link so people browsing can start up a binder of the library and immediately try out the examples

Minimal change with no impact on users other than permitting people coming across this library for the first time to try it quickly.


bnavigator:
This is the wrong repository for that link

jonititan:
Apologies I thought it was going to python-control

Issue #177, opened 2022-06-30
Cannot build or install slycot on Ubuntu


matthieuvigne:
I'm not able to get slycot to install on Ubuntu 22.04. 

This can be reproduced in docker: from a vanilla ubuntu, I install all dependancies then try to perform a pip install:

```
docker container run -it ubuntu bash
```

```
apt update ; apt install python3 python3-dev gcc gfortran libopenblas-dev python3-pip cmake
pip install numpy scikit-build
pip install -v slycot
```

But I end up with the following error:

```
  Traceback (most recent call last):
    File "/usr/local/bin/f2py3", line 8, in <module>
      sys.exit(main())
    File "/tmp/pip-build-env-ve213n4q/overlay/local/lib/python3.10/dist-packages/numpy/f2py/f2py2e.py", line 704, in main
      run_main(sys.argv[1:])
    File "/tmp/pip-build-env-ve213n4q/overlay/local/lib/python3.10/dist-packages/numpy/f2py/f2py2e.py", line 441, in run_main
      postlist = callcrackfortran(files, options)
    File "/tmp/pip-build-env-ve213n4q/overlay/local/lib/python3.10/dist-packages/numpy/f2py/f2py2e.py", line 342, in callcrackfortran
      postlist = crackfortran.crackfortran(files)
    File "/tmp/pip-build-env-ve213n4q/overlay/local/lib/python3.10/dist-packages/numpy/f2py/crackfortran.py", line 3265, in crackfortran
      postlist = postcrack(grouplist[0])
    File "/tmp/pip-build-env-ve213n4q/overlay/local/lib/python3.10/dist-packages/numpy/f2py/crackfortran.py", line 1967, in postcrack
      g = postcrack(g, tab=tab + '\t')
    File "/tmp/pip-build-env-ve213n4q/overlay/local/lib/python3.10/dist-packages/numpy/f2py/crackfortran.py", line 1986, in postcrack
      block['body'] = analyzebody(block, args, tab=tab)
    File "/tmp/pip-build-env-ve213n4q/overlay/local/lib/python3.10/dist-packages/numpy/f2py/crackfortran.py", line 2150, in analyzebody
      not b['body'] and not b['implementedby']:
  KeyError: 'implementedby'
  gmake[2]: *** [slycot/CMakeFiles/_wrapper.dir/build.make:79: slycot/_wrappermodule.c] Error 1
[...]

PR #178, opened 2022-06-30
Work around numpy/numpy#21889

bnavigator:
Works around numpy/numpy#21889 as reported in #177

### DIFF
--- a/.github/conda-env/build-env.yml
+++ b/.github/conda-env/build-env.yml
@@ -1,4 +1,4 @@
 name: build-env
 dependencies:
   - boa
-  - numpy
+  - numpy!=1.23.0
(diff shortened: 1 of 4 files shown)


matthieuvigne:
Looks like a nice workaround for now, until numpy is fixed!

I think you should tag & publish it on pip to have it working again. As is, I tried the install on ubuntu and fedora, and tried installing older versions of slycot - but obviously nothing worked since the issue was with numpy all along :)

bnavigator:
Thanks @matthieuvigne 

Yes, we need a new release. It's been over two years since 4.0.0 already and we made quite some fixes and enhancements.

I don't have access to the pypi credentials thoug. Any thoughts @roryyorke?

roryyorke:
No objections from me. I'll have some time Sunday or next weekend.

roryyorke:
Ran out of time today, sorry. Should have some time next weekend.

PR #188, opened 2022-12-18
pin numpy to <1.24 during build due to gh-187

bnavigator:
Work around for #187


bnavigator:
The random test failures are unrelated (python-control/python-control#818)

PR #190, opened 2022-12-25
update workflow to pass jobname to python-control pytest

murrayrm:
This PR passes information about the current OS and BLAS version that is being used in slycot testing to python-control unit tests to avoid generating errors on known failure modes.  This allows the changes in [python-control PR #821](https://github.com/python-control/python-control/pull/821) to xfail on configurations that were causing errors (outside of slycot functionality).

A better fix would be to somehow set up the python-control unit tests to only run those that rely on slycot functions when using for slycot testing.  But I couldn't think of an easy way to do that.

### DIFF
--- a/.github/workflows/slycot-build-and-test.yml
+++ b/.github/workflows/slycot-build-and-test.yml
@@ -283,7 +283,7 @@ jobs:
           pip install slycot-wheels/${{ matrix.packagekey }}/slycot*.whl
           pip show slycot
       - name: Slycot and python-control tests
-        run: JOBNAME=$JOBNAME bash slycot-src/.github/scripts/run-tests.sh
+        run: JOBNAME="$JOBNAME" bash slycot-src/.github/scripts/run-tests.sh
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/slycot-build-and-test.yml
```diff
@@ -283,7 +283,9 @@ jobs:
           pip install slycot-wheels/${{ matrix.packagekey }}/slycot*.whl
           pip show slycot
       - name: Slycot and python-control tests
-        run: bash slycot-src/.github/scripts/run-tests.sh
+        run: JOBNAME=$JOBNAME bash slycot-src/.github/scripts/run-tests.sh
```
bnavigator, line 286: I am surprised that this works without quotes around `$JOBNAME` as there is a space in the value.

bnavigator:
I noticed that `run-tests.sh` measures and reports the coverage of Slycot. That should be removed here. At least the upload to Coveralls.

murrayrm:
Added quotes just to be safe + added the build type (pip/wheel vs conda) so that we only xfail on conda.  This PR will generate errors until python-control master is updated to check the build type (coming soon).

PR #201, opened 2023-07-20
Add ab04md

KybernetikJo:
Add ab04md

TODO:
- [x] add docstring
- [x] add unit tests
 - [x] check behavior: intent(in, out, copy, ...), see #200 

```python
import slycot
import numpy as np

A = np.array([[1.0, 0.5],
              [0.5, 1.0]])
B = np.array([[0.0, -1.0],
              [1.0, 0.0]])
C = np.array([[-1.0, 0.0],
              [0.0, 1.0]])
D = np.array([[1.0, 0.0],
              [0.0, -1.0]])

Ad = np.array([[-1.0, -4.0],
               [-4.0, -1.0]])
Bd = np.array([[2.8284, 0.0],
               [0.0, -2.8284]])
Cd = np.array([[0.0, 2.8284],
               [-2.8284, 0.0]])
Dd = np.array([[-1.0, 0.0],
               [0.0, -3.0]])

n = 2
m = 2
p = 2

Ad_t, Bd_t, Cd_t, Dd_t = slycot.ab04md('C',n,m,p,A,B,C,D)

print("Ad_t = \n", Ad_t)
print("Bd_t = \n", Bd_t)
print("Cd_t = \n", Cd_t)
print("Dd_t = \n", Dd_t)

Ac_t, Bc_t, Cc_t, Dc_t = slycot.ab04md('D',n,m,p,Ad_t,Bd_t,Cd_t,Dd_t)

print("Ac_t = \n", Ac_t)
print("Bc_t = \n", Bc_t)
print("Cc_t = \n", Cc_t)
print("Dc_t = \n", Dc_t)

print(np.allclose(A,Ac_t))
print(np.allclose(B,Bc_t))
print(np.allclose(C,Cc_t))
print(np.allclose(D,Dc_t))
```

output
```
Ad_t = 
 [[-1. -4.]
 [-4. -1.]]
Bd_t = 
 [[-2.82842712 -0.        ]
 [-0.          2.82842712]]
Cd_t = 
 [[-0.          2.82842712]
 [-2.82842712  0.        ]]
Dd_t = 
 [[3. 0.]
 [0. 1.]]
Ac_t = 
 [[1.  0.5]
 [0.5 1. ]]
Bc_t = 
 [[ 0. -1.]
 [ 1. -0.]]
Cc_t = 
 [[-1.  0.]
 [-0.  1.]]
Dc_t = 
 [[ 1.  0.]
 [ 0. -1.]]
True
True
True
True
```

PR #207, opened 2023-08-25
Replace unittest calls with pytest

bnavigator:
Unify the test suite api with Pytest; remove any remnants of old unittest usage.


bnavigator:
This also fixes an error I made as a last minute addition in #201, which made the test not discoverable for pytest

PR #213, opened 2023-08-27
Change transform.py to numpydoc style

KybernetikJo:
This PR refactors the `Slycot/slycot/transform.py` in order to be closer to the numpydoc style as discussed in #100.
Should be part of #204.

(finished 12 of 12 routines)

| Routine name| Finished |Status  description  |
|---------------------|--------------|-----------------------------|
| tb01id  | <ul><li>[x] </li></ul> | *3 |
| tb10pd  | <ul><li>[x] </li></ul> | *3 |
| tb03ad  | <ul><li>[x] </li></ul> | *3 |
| tb04ad  | <ul><li>[x] </li></ul> | *3 |
| tb05ad | <ul><li>[x] </li></ul> | *3 |
| tc01od  | <ul><li>[x] </li></ul> | *3 |
| tc04ad  | <ul><li>[x] </li></ul> | *3 |
| td04ad  | <ul><li>[x] </li></ul> | *3 |
| tf01md  | <ul><li>[x] </li></ul> | *3 |
| tf01rd  | <ul><li>[x] </li></ul> | *3 |
| tf01rd  | <ul><li>[x] </li></ul> | *3 |
| tf01rd  | <ul><li>[x] </li></ul> | *3 |

Status description:
1. numpydoc docstring sections, fixed
2. parameters data types, fixed and 1 fixed
3. optional and default values, fixed and 2 fixed

PR #212, opened 2023-08-27
Change synthesis.py to numpydoc style

KybernetikJo:
This PR refactors the `Slycot/slycot/synthesis.py` in order to be closer to the numpydoc style as discussed in #100.
Should be part of #204.

Remark:
The synthesis.py has already been quite close to numpydoc before this PR. There must have been major cleanups in the past.

(finished 18 of 18 = (17 + 1) routines)

| Routine name| Finished |Status  description  |
|---------------------|--------------|-----------------------------|
| sb01bd  | <ul><li>[x] </li></ul> | *3 |
| sb02md  | <ul><li>[x] </li></ul> | *3  |
| sb02mt  | <ul><li>[x] </li></ul> | *3 |
| sb02od  | <ul><li>[x] </li></ul> | *3 |
| sb03md | <ul><li>[x] </li></ul> | *3 |
| sb03md57  | <ul><li>[x] </li></ul> | *3  |
| sb03od  | <ul><li>[x] </li></ul> | *3 |
| sb04md  | <ul><li>[x] </li></ul> | *3 |
| sb04qd  | <ul><li>[x] </li></ul> | *3 |
| sb10ad  | <ul><li>[x] </li></ul> | *3 |
| sb10dd  | <ul><li>[x] </li></ul> | *3 |
| sb10fd  | <ul><li>[x] </li></ul> | *3 |
| sb10hd  | <ul><li>[x] </li></ul> | *3 |
| sb10jd  | <ul><li>[x] </li></ul> | *3 |
| sb10yd  | <ul><li>[x] </li></ul> | *3 |
| sg02ad  | <ul><li>[x] </li></ul> | *3 |
| sg03ad  | <ul><li>[x] </li></ul> | *3 |
| sg03bd  | <ul><li>[x] </li></ul> | *3 |

Status description:
1. numpydoc docstring sections, fixed
2. parameters data types, fixed and 1 fixed
3. optional and default values, fixed and 2 fixed

PR #211, opened 2023-08-27
Change math.py to numpydoc style

KybernetikJo:
This PR refactors the `Slycot/slycot/math.py` in order to be closer to the numpdoc style as discussed in #100.
Should be part of #204.

(finished 7 of 7 routines)

| Routine name| Finished |Status  description  |
|---------------------|--------------|-----------------------------|
| mb03rd  | <ul><li>[x] </li></ul> | *3 |
| mb03vd  | <ul><li>[x] </li></ul> | *3 |
| mb03vy  | <ul><li>[x] </li></ul> | *3 |
| mb03wd  | <ul><li>[x] </li></ul> | *3 |
| mb05md | <ul><li>[x] </li></ul> | *3 |
| mb05nd  | <ul><li>[x] </li></ul> | *3 |
| mc01td  | <ul><li>[x] </li></ul> | *3 |

Status description:
1. numpydoc docstring sections, fixed
2. parameters data types, fixed and 1 fixed
3. optional and default values, fixed and 2 fixed

PR #206, opened 2023-08-22
Change analysis.py to numpydoc style

KybernetikJo:
This PR refactors the `Slycot/slycot/analysis.py` in order to be closer to the numpydoc style as discussed in #100.
Should be part of #204.

(finished 0 of 18 routines)

| Routine name| Finished |Status  description  |
|---------------------|--------------|-----------------------------|
| ab01nd  | <ul><li>[ ] </li></ul> | *3 |
| ab05md  | <ul><li>[ ] </li></ul> | *3   |
| ab05nd  | <ul><li>[ ] </li></ul> |  *3 |
| ab07nd  | <ul><li>[ ] </li></ul> |  *3 |
| ab08nd  | <ul><li>[ ] </li></ul> |  *3 |
| ab08nz  | <ul><li>[ ] </li></ul> |  *3 |
| ab09ad  | <ul><li>[ ] </li></ul> |  *3 |
| ab09ax  | <ul><li>[ ] </li></ul> |  *3 |
| ab09bd  | <ul><li>[ ] </li></ul> |  *3 |
| ab09md  | <ul><li>[ ] </li></ul> |  *3 |
| ab09nd  | <ul><li>[ ] </li></ul> |  *3 |
| ab13bd  | <ul><li>[ ] </li></ul> |  *3 |
| ab13dd  | <ul><li>[ ] </li></ul> |  *3 |
| ab13ed  | <ul><li>[ ] </li></ul> |  *3 |
| ab13fd  | <ul><li>[ ] </li></ul> |  *3 |
| ab13md  | <ul><li>[ ] </li></ul> |  *3 |
| ag08bd  | <ul><li>[ ] </li></ul> |  *3 |

Status description:
1. numpydoc docstring sections, fixed
2. parameters data types, fixed and 1 fixed
3. optional and default values, fixed and 2 fixed
4. ...., fixed and 3 fixed

At least every slycot routine should have the following [numpydoc docstring sections](https://numpydoc.readthedocs.io/en/latest/format.html#sections)
- Parameters
- Returns
- Raises or Warns

Scipy often uses the numpydoc docstring sections 
- Parameters
- Returns
- See Also
- Notes
- References
- Examples
[...]

PR #225, opened 2023-12-29
Add ab09hd and tg01gd, and show that n,m,p can be removed

mccullerlp:
Hi all,
I wanted to test using ab09hd as it can ensure relative error during reduction. Additionally, tg01gd is useful for maintaining descriptor systems and reducing them back to a standard form. I've added wrappers and python interfaces. It's less work if you are willing to pull directly, but I could split this into separate PR's. The docstrings are possibly too much of a copy of the slicot documentation?

Additionally, I added an auxiliary wrapper for ab09nd that adjusts the dependency resolution between n,m,p and a,b,c. This is a general technique that can allow the n,m,p arguments to be dropped from any/all of the wrappers. I know this changes the interface, since n,m,p are exposed to the external API, even though they could all be determined from the python-side wrappers anyway. Including it since I already tested it, and it may be of some interest to simplify the interfaces.


bnavigator:
Thank you very much for your enthusiasm and work on Slycot.

- Please rebase your contribution onto current master (c2117e2863051eae8dbd12b3e1393f8a481675ae), so that merge conflicts are resolved and we can run the CI.
- Please adjust the docstring to the numpydoc standard, see @KybernetikJo's open Pull requests  #206, #211, #212, and #213 as examples.
- New wrappers can leverage calculating parameters implicitly, some do already, some really old ones don't. Whenever I made a new wrapper it was a hit an miss for me though about what f2py could deduce in the right order and allocate correctly. For this reason, it's mandatory that you create meaningful unit tests.

.gitignore
```diff
@@ -10,5 +10,7 @@ build.log
 *.egg-info/
 .coverage
 *~
+.#*
+.*~
```
bnavigator, line 14: `.*~` This is already covered by `*~`
What kind of tool creates files with `#` in it?

slycot/transform.py
```diff
@@ -1236,4 +1236,245 @@ def tg01fd(l,n,m,p,A,E,B,C,Q=None,Z=None,compq='N',compz='N',joba='N',tol=0.0,ld
...
+
+    out = _wrapper.tg01gd(jobs,l,n,m,p,a,e,b,c,d,tol)
+    info = out[-1]
+    raise_if_slycot_error(info, arg_list)
+    A,E,B,C,D,lr,nr,ranke,infred,info = out
+    return (
+        A[..., :nr, :nr],
+        E[..., :nr, :nr],
+        B[..., :nr, :],
+        C[..., :, :nr],
```
bnavigator, line 1472: Remove the leading `...` parts. The wrapper only returns two-dimensional arrays.

bnavigator:
Hi @mccullerlp, glad to see you working on this branch. My comments from December 2023 are still valid. Please make a rebase. There are lots of changes [listed in this PR diff](https://github.com/python-control/Slycot/pull/225/files] that are unrelated.

PR #237, opened 2024-09-04
Update macos gfortran version

slivingston:
Entirely similar to https://github.com/python-control/python-control/pull/1041


bnavigator:
Thanks!

PR #240, opened 2024-11-11
Constrain NumPy version to 1.X

sdahdah:
It seems that the Slycot wheels on PyPI were built with NumPy 1.X, but the `pyproject.toml` allows NumPy 2.X to be installed.

This means that if I create a fresh virtual environment and install Slycot, then it will pull NumPy 2.X, which makes the Slycot import fail.

I've added the version constraint in this PR, but if this is not what you had in mind, feel free to just close without merging.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = [
     "wheel",
     "scikit-build>=0.15",
     "cmake>=3.14",
-    "numpy >= 1.23.1"]
+    "numpy>=1.23.1,<2"]
 build-backend = "setuptools.build_meta"
 
 [project]
@@ -37,7 +37,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-    "numpy >= 1.23.1",
+    "numpy>=1.23.1,<2",
 ]
 dynamic = ["version"]
 


slivingston:
The `control` package is intended to support NumPy 2 (https://github.com/python-control/python-control/pull/1013 and https://github.com/python-control/python-control/pull/994), and `control` uses Slycot, so the key question is: can Slycot be made compatible with both NumPy 1 and 2?

If not, or if it requires a lot of work, then your proposed change is a good temporary solution.

sdahdah:
I think building the wheels with NumPy 2 makes the resulting package compatible with NumPy 1 and 2

bnavigator:
There are no Slycot wheels on PyPI. We only publish the source distribution and it is compatible with NumPy 2.

sdahdah:
> There are no Slycot wheels on PyPI. We only publish the source distribution and it is compatible with NumPy 2.

My bad, there must be something else wrong on my end.

bnavigator:
No worries, feel free to ask if you need more help. You probably just have to recompile/reinstall Slycot with numpy 2. Make sure to not use a previously compiled wheel from your local cache.

sdahdah:
Thanks, the problem was actually a previously compiled wheel in my cache.

PR #247, opened 2025-02-17
CI: skip docstrings_test.py and add defaults channel

slivingston:
`numpydoc` recently became required [in docstrings_test.py of python-control](https://github.com/python-control/python-control/blob/d11f05dbadd8218b5d8285d32115f1841e3a2ce5/control/tests/docstrings_test.py#L23). Slycot CI jobs include python-control tests [and skip irrelevant ones](https://github.com/python-control/Slycot/blob/bbfe718e5e32934ac64c0ac90ff08a2d18c7f389/.github/scripts/run-tests.sh#L14-L21). This PR skips docstrings_test.py because docstring consistency tests of python-control are not relevant here, and doing so also avoids the need to install numpydoc in these CI jobs.

This PR also add explicit miniconda channels as in
* https://github.com/python-control/python-control/pull/1129
* https://github.com/python-control/python-control/pull/1128

### DIFF
--- a/.github/scripts/run-tests.sh
+++ b/.github/scripts/run-tests.sh
@@ -18,6 +18,7 @@ donttest="$donttest or test_default_deprecation"
 pytest control/tests \
         --cov=$slycot_libdir \
         --cov-config=${slycot_srcdir}/.coveragerc \
+        --ignore=control/tests/docstrings_test.py \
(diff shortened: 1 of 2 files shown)


bnavigator (approved):

bnavigator:
Thanks! Makes sense.

PR #250, opened 2025-09-07
Get conda builds in GIthub workflow mostly working

roryyorke:
meta.yaml and conda_build_config.yaml copied and adapted from conda-forge/slycot-feedstock.

The failures in testing on OSX seem to be due to something with coveralls; the failures happen (at least on my repo - tests yet to run here) at coverage-collection time, and this happens for both conda and wheel tests.  I didn't change the wheel build, and these OSX wheel tests passed completely in the most recent merged PR gh-294.

I tried to get rid of the various conda warnings encountered through the build, but ended up with a free-threaded python (!) when testing against 3.13; when I tried to change `meta.yaml` to prevent that, the build broke in a mysterious way.

### DIFF
--- a/.github/conda-env/build-env.yml
+++ b/.github/conda-env/build-env.yml
@@ -1,4 +1,0 @@
-name: build-env
-dependencies:
-  - boa
-  - numpy
--- a/.github/conda-env/test-env.yml
+++ b/.github/conda-env/test-env.yml
@@ -1,0 +1,9 @@
+name: test-env
+dependencies:
+  # in addtion to package dependencies and explicit LAPACK/BLAS implementations installed in workflow
+  - scipy
+  - matplotlib
+  - pytest
+  - pytest-cov
+  - pytest-timeout
+  - coverage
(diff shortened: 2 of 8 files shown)


roryyorke:
I've opened python-control/python-control#1182 to allow only Slycot-dependent tests to be selected in pytest.

roryyorke:
Tests pass, at last.  I haven't checked the coverage decrease, but it may be because of decrease in coverage of python-control, not Slycot, since this only runs Slycot-related python-control tests.  The reported decrease is relatively modest because I think `coverage` only reports coverage of Python files that are actually run. I imagine the true coverage decrease in python-control is higher.  That's fine: we want to test Slycot-related things, not the whole of python-control.

I see `auto-update-conda` was previously explicitly set to `false`, I've set it to `true`; I did this because of the CI failures in https://github.com/python-control/Slycot/pull/250/commits/2c899321171d2807fce0a89fd80374d7ac24c0cb, which I'm pretty sure have nothing to do with Slycot.  I don't know why it was previously set to false.

roryyorke:
Coverage restored.  Any objections to merging this?

